PR #3, opened 2014-12-18
Added services examples

esteve:
This branch adds service and client examples.

### DIFF
--- a/userland/src/add_two_ints_client.cpp
+++ b/userland/src/add_two_ints_client.cpp
@@ -30,15 +30,20 @@ int main(int argc, char** argv)
   request->a = 2;
   request->b = 3;
 
-  auto rc = client->send_request(request, response);
-  if(rc == 0)
+  auto f = client->async_send_request(request, response);
+
+  std::future_status status;
+  do
   {
-    std::cout << "Sum: " << response->sum << std::endl;
-    return 0;
-  } else
+    rclcpp::spin_some(node);
+    status = f.wait_for(std::chrono::milliseconds(100));
+  } while(status != std::future_status::ready && rclcpp::ok());
+
+  if(std::future_status::ready == status)
   {
-    std::cerr << "Error receiving a response" << std::endl;
+    std::cout << "FUTURE READY" << std::endl;
+    std::cout << f.get()->sum << std::endl;
   }
 
-  return rc;
+  return 0;
 }
--- a/userland/src/add_two_ints_client.cpp
+++ b/userland/src/add_two_ints_client.cpp
@@ -18,7 +18,6 @@
 
 #include <userland_msgs/AddTwoInts.h>
 
-
 int main(int argc, char** argv)
 {
   rclcpp::init(argc, argv);
@@ -26,21 +25,19 @@ int main(int argc, char** argv)
   auto node = rclcpp::Node::make_shared("add_two_ints_client");
 
   auto client = node->create_client<userland_msgs::AddTwoInts>("add_two_ints");
-  auto req = std::make_shared<userland_msgs::AddTwoInts::Request>();
-  req->a = 2;
-  req->b = 3;
+  auto request = std::make_shared<userland_msgs::AddTwoInts::Request>();
+  request->a = 2;
+  request->b = 3;
 
-  std::shared_ptr<userland_msgs::AddTwoInts::Response> response = client->send_request(req);
-  std::cout << "Sum: " << response->sum << std::endl;
-
-  response = client->send_request(req);
-  std::cout << "Sum: " << response->sum << std::endl;
-
-  response = client->send_request(req);
-  std::cout << "Sum: " << response->sum << std::endl;
-
-  response = client->send_request(req);
-  std::cout << "Sum: " << response->sum << std::endl;
+  try
+  {
+    auto response = client->send_request(request);
+    std::cout << "Sum: " << response->sum << std::endl;
+  } catch(const std::runtime_error &e)
+  {
+    std::cerr << "Timed out while waiting for a response" << std::endl;
+    return 1;
+  }
 
   return 0;
 }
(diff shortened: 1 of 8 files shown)


userland/src/add_two_ints_server.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <simple_msgs/AllPrimitiveTypes.h>
+#include <simple_msgs/AllStaticArrayTypes.h>
+#include <simple_msgs/Nested.h>
+#include <simple_msgs/String.h>
+#include <simple_msgs/Uint32.h>
+
+#include <userland_msgs/AddTwoInts.h>
+#include <userland_msgs/RequestId.h>
+
+void add(const std::shared_ptr<userland_msgs::AddTwoIntsRequest> req,
```
dirk-thomas: Use `userland_msgs::AddTwoInts::Request` here.

userland/src/add_two_ints_server.cpp
```diff
@@ -0,0 +1,42 @@
...
+         const std::shared_ptr<userland_msgs::RequestId> req_id,
+         std::shared_ptr<userland_msgs::AddTwoIntsResponse> res)
```
dirk-thomas: Use `userland_msgs::AddTwoInts::Response` here.

userland/src/add_two_ints_server.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <simple_msgs/AllStaticArrayTypes.h>
+#include <simple_msgs/Nested.h>
+#include <simple_msgs/String.h>
+#include <simple_msgs/Uint32.h>
+
+#include <userland_msgs/AddTwoInts.h>
+#include <userland_msgs/RequestId.h>
+
+void add(const std::shared_ptr<userland_msgs::AddTwoIntsRequest> req,
+         const std::shared_ptr<userland_msgs::RequestId> req_id,
```
dirk-thomas: The `RequestId` should come from a "RPC-namespace" instead of the message package.

userland/src/add_two_ints_client.cpp
```diff
@@ -0,0 +1,43 @@
...
+  auto node = rclcpp::Node::make_shared("add_two_ints_client");
+
+  auto client = node->create_client<userland_msgs::AddTwoInts>("add_two_ints");
+  auto request = std::make_shared<userland_msgs::AddTwoInts::Request>();
+  request->a = 2;
+  request->b = 3;
+
+  try
+  {
+    auto response = client->send_request(request);
```
dirk-thomas: Since this signature implies allocation of the response we should implement a different signature on the lower level:

    rc = client->send_request(request, response);

This way the response can be preallocated by the user land code.
The previous signature can be provided as syntactic sugar on-top.

userland/src/add_two_ints_client.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <userland_msgs/AddTwoInts.h>
+
+int main(int argc, char** argv)
+{
+  rclcpp::init(argc, argv);
+
+  auto node = rclcpp::Node::make_shared("add_two_ints_client");
+
+  auto client = node->create_client<userland_msgs::AddTwoInts>("add_two_ints");
```
dirk-thomas: Adding an asynchronous example (beside this synchronous one) would be good in the future.

esteve: I modified the example to work asynchronously. Unfortunately, I can't hide the do { ... } loop because I can't spin the from inside the client.

wjwwood: I was expecting something like this for the asynchronous case:

```c++
void handle_response(const std::shared_ptr<userland_msgs::AddTwoInts::Response> &response)
{
  std::cout << "Got response: " << response->sum << std::endl;
}

...
client->async_send_request(request, handle_response);  // I guess you pass a pre-allocated response too
rclcpp::spin();  // This could be here or in another thread and would cause the callback to get called
```

wjwwood: I think the future method supports other use cases as well though, we may need something similar to both.

Even in the future method as you have here I would have expected the spin to be elsewhere, though not necessarily.

esteve: I expected to be able to spin inside the client, but alas the API doesn't allow that, or I couldn't figure out how. Where else did you expect to spin?

wjwwood: We talked about the issues of spinning inside of a client, you'll need a recursive lock and a way to deal with multithreaded execution...

You can spin in different threads, for example you could create thread to call spin in or you could create the async request in a callback for subscription or you could spin in the main thread and then do the async request in a user created thread. There are lots of options.

The callback pattern that I described above has the advantage that you don't have to poll the future, so its more like request it and forget it. The future pattern has the advantage that you can use one thread to do work while periodically checking the future. But in both cases you need to be spinning in the mean time. In your example you just happen to use the single thread for polling the future and spinning.

wjwwood: Either way this looks like a good proof of concept, we can certainly implement what I described above with the existing functionality. I think we should get a consensus on the API before we spend more time iterating on more API features. To that end I think we should get these branches merged and then we should have a meeting to sync up on the API's, I'll try to setup a meeting today.

wjwwood:
+1 for me.

esteve:
@wjwwood This branch depends on RTI's messaging library, which we haven't packaged and for which we don't have the source code, so this will break the current code unless we all install the messaging library. In any case, it'd be relatively straightfoward to reimplement that library on top of the existing classes or just implement the spec.

wjwwood:
Ah, I forgot about needing to synchronize with the other packages.

I am +1'ing the current examples and rclcpp API. If we are blocked on the implementations for specific DDS vendors then in the mean time we can raise not implemented errors at compile time or run time. by doing this we can at least continue to iterate on the user facing API and the middleware API with your service changes in place.

Getting it working with RTI's public version and with OpenSplice is important, but we shouldn't let stalls there stop our work in trying to get our API's figured out.

esteve:
Ping @tfoote @dirk-thomas

tfoote:
+1 for the API

PR #6, opened 2015-03-25
remove include_directories() call in msgs packages

wjwwood:
This pull request removes several calls to `include_directories()` in the message packages, which added the include dirs of the `rosidl_default_generators` package, which was therefore causing the generated code to include the wrong version of the `rosidl_generator_cpp/get_type_support_handle_impl.hpp` header.

Connects to ros2/ros2#26


esteve:
+1

tfoote:
+1

dirk-thomas:
+1

PR #14, opened 2015-04-17
Added example for the parameters API

esteve:
This PR adds an example of the developer API for parameters.

@dirk-thomas @tfoote @wjwwood 

Connects to ros2/ros2#28

Depends on ros2/rclcpp#19
Depends on ros2/rclcpp#20

### DIFF
--- a/userland/src/parameters.cpp
+++ b/userland/src/parameters.cpp
@@ -7,7 +7,7 @@
 
 #include <rclcpp/rclcpp.hpp>
 
-template <typename ParameterTypeT>
+template<typename ParameterTypeT>
 ParameterTypeT
 wait_for_future(rclcpp::node::Node::SharedPtr & node, std::shared_future<ParameterTypeT> & future)
 {
@@ -24,7 +24,7 @@ wait_for_future(rclcpp::node::Node::SharedPtr & node, std::shared_future<Paramet
   throw std::runtime_error("Invalid future status");
 }
 
-int main(int argc, char** argv)
+int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
 
@@ -46,7 +46,8 @@ int main(int argc, char** argv)
   std::cout << "Parameter (foo) found: " << found << std::endl;
 
   const std::vector<rclcpp::parameter::ParameterContainer> kv = {
-    rclcpp::parameter::ParameterContainer("foo", value1)};
+    rclcpp::parameter::ParameterContainer("foo", value1)
+  };
 
   auto f4 = node->async_set_parameters(node_name, kv);
   wait_for_future<bool>(node, f4);
@@ -55,18 +56,19 @@ int main(int argc, char** argv)
   int64_t value2 = wait_for_future<int64_t>(node, f5);
   std::cout << "Value for parameter (foo): " << value2 << std::endl;
 
-  const std::vector<std::string> pn = { "foo" };
+  const std::vector<std::string> pn = {"foo"};
   auto f6 = node->async_get_parameters(node_name, pn);
-  auto multi_value = wait_for_future< std::vector<rclcpp::parameter::ParameterContainer> >(node, f6);
+  auto multi_value = wait_for_future<std::vector<rclcpp::parameter::ParameterContainer>>(node, f6);
 
-  for(auto v : multi_value) {
+  for (auto v : multi_value) {
     std::cout << "Value for parameter (foo) (multi value): " << v.get_value<int64_t>() << std::endl;
   }
 
   auto f7 = node->async_get_parameter<int64_t>(
-    node_name, "foo", [] (std::shared_future<int64_t> f) {
-      std::cout << "Value for parameter (foo) (callback): " << f.get() << std::endl; }
-  );
+    node_name, "foo", [](std::shared_future<int64_t> f) {
+    std::cout << "Value for parameter (foo) (callback): " << f.get() << std::endl;
+  }
+    );
   wait_for_future<int64_t>(node, f7);
   rclcpp::spin(node);
 
--- a/userland/src/parameters.cpp
+++ b/userland/src/parameters.cpp
@@ -1,0 +1,74 @@
+#include <algorithm>
+#include <chrono>
+#include <cmath>
+#include <iostream>
+#include <string>
+#include <thread>
+
+#include <rclcpp/rclcpp.hpp>
+
+template <typename ParameterTypeT>
+ParameterTypeT
+wait_for_future(rclcpp::node::Node::SharedPtr & node, std::shared_future<ParameterTypeT> & future)
+{
+  std::future_status status;
+  do {
+    rclcpp::spin_some(node);
+    status = future.wait_for(std::chrono::milliseconds(100));
+  } while (status != std::future_status::ready && rclcpp::ok());
+
+  if (std::future_status::ready == status) {
+    return future.get();
+  }
+
+  throw std::runtime_error("Invalid future status");
+}
+
+int main(int argc, char** argv)
+{
+  rclcpp::init(argc, argv);
+
+  const std::string node_name("paramers_example");
+
+  rclcpp::node::Node::SharedPtr node = rclcpp::Node::make_shared(node_name);
+
+  auto f1 = node->async_has_parameter(node_name, rclcpp::parameter::ParameterQuery("foo"));
+  bool found = wait_for_future<bool>(node, f1);
+  std::cout << "Parameter (foo) found: " << found << std::endl;
+
+  int64_t value1 = 1234;
+  auto f2 = node->async_set_parameter(node_name, "foo", value1);
+  wait_for_future<bool>(node, f2);
+  std::cout << "Setting value: " << value1 << " for parameter (foo)" << std::endl;
+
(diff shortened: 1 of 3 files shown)


userland/test/CMakeLists.txt
```diff
@@ -4,12 +4,14 @@ foreach(middleware_impl_tmp ${middleware_implementations})
 
   add_executable(test_publisher__${middleware_impl_tmp} "test_publisher.cpp")
   ament_target_dependencies(test_publisher__${middleware_impl_tmp}
+    "rcl_interfaces"
```
dirk-thomas: Why does this need to declare a dependency on `rcl_interfaces`? It is not using it directly. If it is a transitive requirement for `rclcpp` then `rclcpp` should export a dependency on `rcl_interfaces`.

esteve: Agreed, but adding `rcl_interfaces` as a dependency to `rclcpp`'s `package.xml` does not seem to change anything. Is there any other place I should change?

dirk-thomas: `rclcpp` needs to also export the dependency in CMake using `ament_export_dependencies`.

esteve: Thanks. I just updated ros2/rclcpp#20: https://github.com/ros2/rclcpp/commit/f4437b3d4574fe61db5c95adc4051aea9c5bd865

dirk-thomas: `rclcpp` still lacks a run dependency in the package.xml, e.g. a `build_export_depend`.

userland/src/parameters.cpp
```diff
@@ -0,0 +1,74 @@
...
+    rclcpp::spin_some(node);
+    status = future.wait_for(std::chrono::milliseconds(100));
+  } while (status != std::future_status::ready && rclcpp::ok());
+
+  if (std::future_status::ready == status) {
+    return future.get();
+  }
+
+  throw std::runtime_error("Invalid future status");
+}
```
wjwwood: This seems pretty boiler plate, maybe we should provide this functionality for the user, either as a method of one of our classes that takes a future, or instead of returning a std future object return a custom future object which wraps it and provides this functionality.

userland/src/parameters.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <rclcpp/rclcpp.hpp>
+
+template <typename ParameterTypeT>
+ParameterTypeT
+wait_for_future(rclcpp::node::Node::SharedPtr & node, std::shared_future<ParameterTypeT> & future)
+{
+  std::future_status status;
+  do {
+    rclcpp::spin_some(node);
+    status = future.wait_for(std::chrono::milliseconds(100));
```
wjwwood: What's the rational for waiting here? Wouldn't it be better to wait for something to happen in the spin and then do a non-blocking check and then spin again until the future is ready? It seems to me that the future will never become ready except during a spin, so waiting outside of the spin doesn't do anything other than slow down the busy loop polling of `spin_some`.

Instead we should get a blocking version of `spin_some` which can block until something happens and the do a non-blocking check of the future's status.

userland/src/parameters.cpp
```diff
@@ -0,0 +1,74 @@
...
+  throw std::runtime_error("Invalid future status");
+}
+
+int main(int argc, char** argv)
+{
+  rclcpp::init(argc, argv);
+
+  const std::string node_name("paramers_example");
+
+  rclcpp::node::Node::SharedPtr node = rclcpp::Node::make_shared(node_name);
```
wjwwood: I think we should try to keep the examples concise, I would have done this:

```c++
auto node = rclcpp::Node::make_shared("parameters_example");
```

Why expand it out with a separate variable for the node name? I guess you just want to be explicit?

userland/src/parameters.cpp
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  throw std::runtime_error("Invalid future status");
+}
+
+int main(int argc, char** argv)
+{
+  rclcpp::init(argc, argv);
+
+  const std::string node_name("paramers_example");
```
wjwwood: `"parameters_example"`

dirk-thomas:
Should we also start adding unit tests for new features with the jobs finally turning over?

userland/src/parameters.cpp
esteve: +1 to the blocking version of `spin_some`

userland/src/parameters.cpp
esteve: I'd rather add this to the synchronous version of the client than rolling our own future or extending `std::future`/`std::shared_future`

esteve:
@dirk-thomas +1 to adding tests

userland/src/parameters.cpp
wjwwood: > I'd rather add this to the synchronous version of the client

Can you explain in more detail what you mean by adding it to the synchronous version? As a function or a method?

The users using the async API also need this, consider:

```c++
auto f = node->async_has_parameter(...);
// User does some work in the mean time
bool found = wait_for_future<bool>(node, f);
```

I think this function provides utility that is common enough that it should be included with the standard async API.

userland/src/parameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+
+  const std::string node_name("parameters_example");
+
+  auto node = rclcpp::Node::make_shared(node_name);
+
+  auto f1 = node->async_has_parameter(node_name, rclcpp::parameter::ParameterQuery("foo"));
```
wjwwood: Is it required that you create a `rclcpp::parameter::ParameterQuery` here? Can you instead just pass a string or a literal?

userland/src/parameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+  bool found = wait_for_future<bool>(node, f1);
+  std::cout << "Parameter (foo) found: " << found << std::endl;
```
wjwwood: Does this print out `0` or `1`? Or does it do something better for readability? Maybe it needs `(found ? "true" : "false")`?

userland/src/parameters.cpp
esteve: I meant hiding this code inside the synchronous calls, so that a synchronous client would return the value held by the future directly.

I'm ok with exposing `wait_for_future` somehow, but not extending `std::future`

userland/src/parameters.cpp
esteve: This code was from before we discussed what `has_parameter` should look like this morning. We concluded that there's no need to check other metadata (e.g. type) when calling `has_parameter`, I'll change the API to just accept a `ParameterName`

userland/src/parameters.cpp
wjwwood: Yeah, I agree about not extending `std::future`, but I think that something like `wait_for_parameter_future` should be part of the async parameter API and the synchronous parameter API should use that function in its implementation.

wjwwood: Or maybe `spin_and_wait_for_future`, since it really is just waiting for a `std::future` which requires spin to get resolved.

wjwwood: Actually Python's asyncio has a `run_until_complete(future)` concept, maybe we can imitate that with something similar, like `spin_until_complete(future)`.

userland/src/parameters.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  int64_t value1 = 1234;
+  auto f2 = node->async_set_parameter(node_name, "foo", value1);
+  wait_for_future<bool>(node, f2);
```
wjwwood: Should your check the return value here?

userland/src/parameters.cpp
wjwwood: Ok, sounds good.

userland/src/parameters.cpp
esteve: `spin_until_complete(future)` looks good to me.

userland/src/parameters.cpp
esteve: The reason for waiting is that you need to wait to know the future's status (ready, deferred, timeout). Here the future here is waiting for a very short time so as to block spinning the least as possible. I can make the timeout even shorter, though.

wjwwood: Looks like you can pass a 0 duration or use `wait_util` and give it a time in the past:

http://stackoverflow.com/a/10917945/671658

Also, they are going to add a `.is_ready()` non-blocking check in the future and there might already be a `._is_ready` function. See:

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3857.pdf (search for `is_ready`)

esteve: Moved this to `rclcpp` and `rclcpp::executors`:

https://github.com/ros2/rclcpp/pull/20/files#diff-690567624ee368716009118718985564R81
https://github.com/ros2/rclcpp/pull/20/files#diff-294cbb523fb3d56a1df2a0a7ff0beb9bR31

wjwwood: Looks ok to me, but in the future at some point we need to make the spin functions recursive safe. For instance, I'm not sure what happens when you call `spin_until_future_complete` within a callback which was executed by another `spin*` call.

PR #13, opened 2015-04-16
add unit tests for server / client

dirk-thomas:
Connects to ros2/ros2#27.

http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/29/

The test passes for:
* Connect cpp
* OpenSplice cpp

It fails for:
* Connext XTypes

since support for services has not been implemented for that rmw implementation yet.

PR #15, opened 2015-04-20
Pass request header to callbacks

esteve:
Connects to ros2/rclcpp#21

### DIFF
--- a/userland/src/add_two_ints_server.cpp
+++ b/userland/src/add_two_ints_server.cpp
@@ -32,12 +32,13 @@
 #include <userland_msgs/AddTwoInts.h>
 
 void add(
-  const std::shared_ptr<userland_msgs::AddTwoInts::Request> req,
-  std::shared_ptr<userland_msgs::AddTwoInts::Response> res)
+  const std::shared_ptr<rmw_request_id_t> request_header,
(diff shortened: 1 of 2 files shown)


esteve:
@dirk-thomas @tfoote @wjwwood

wjwwood:
We don't have to do this now, but in the future it would be nice to make this signature optional. While we should make the header available to those who want it, most people don't care about it and only want the request and response.

wjwwood:
Otherwise, +1

esteve:
@wjwwood I just added support for both types of callbacks (with and without request headers).

esteve:
@wjwwood https://github.com/ros2/rclcpp/commit/8abae8ac829946620c9defd983e1f947839b041a

dirk-thomas:
With #13 being merged please also update the test to match the new API. Also consider using the new `spin_until` functions as commented in https://github.com/ros2/examples/pull/13#discussion_r28722152

dirk-thomas:
+1

esteve:
@dirk-thomas the PR where the `spin_*` functions are implemented is not yet merged because the PR is not reviewed yet :wink:

tfoote:
+1

PR #17, opened 2015-04-28
Added example for the new async parameters API

esteve:
Connects to ros2/ros2#28

@dirk-thomas @tfoote @wjwwood

### DIFF
--- a/userland/src/parameters.cpp
+++ b/userland/src/parameters.cpp
@@ -1,0 +1,104 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


userland/src/parameters.cpp
```diff
@@ -0,0 +1,71 @@
...
+  auto f1 = parameters_client->set_parameters(
+    parameters, [](std::shared_future<std::vector<rcl_interfaces::ParameterSetResult>> f) {
+    for(auto v : f.get()) {
+      if(!v.successful) {
+        std::cerr << v.reason << std::endl;
+      }
+    }
+  });
+
+  rclcpp::spin_until_future_complete<std::vector<rcl_interfaces::ParameterSetResult>>(node, f1);
```
tfoote: Is there any way we can get this to automatically infer the template type?

tfoote:
One question but lgtm

userland/src/parameters.cpp
esteve: I've changed the example to use the inferred template parameter.

tfoote: Nice! That's very clean now.

dirk-thomas:
The async API of `AsyncParametersClient` lgtm.

Should the example also contain some API calls to use the results of the futures to see how a user accesses the data? WIth all the `auto` *magic* it is less clear what to do with `fN` from reading the example.

Nitpicking:
* copyright year
* unused includes
* unnecessary `std::string` for `ParameterVariant` constructor parameters

PR #18, opened 2015-04-29
fix ambiguous overloaded function call on Windows

wjwwood:
Connects to ros2/rmw_connext#26

### DIFF
--- a/userland/src/add_two_ints_server.cpp
+++ b/userland/src/add_two_ints_server.cpp
@@ -47,7 +47,10 @@ int main(int argc, char ** argv)
 
   auto node = rclcpp::Node::make_shared("add_two_ints_server");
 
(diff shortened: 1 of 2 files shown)


wjwwood:
This is basically the patch @esteve proposed, I could not find a way to make it work without this... I tried `explicit` on the constructors and I tried using `auto` to store the function before passing it in, but neither of those worked.

wjwwood:
For review.

dirk-thomas:
Is a similar change also necessary for the tests?

tfoote:
lgtm

esteve:
+1, I have an idea of using std::enable_if and std::is_convertible, but I rather get this merged now and see if we can improve it later.

wjwwood:
Sounds good to me, Thanks.

wjwwood:
I created an issue to track it: https://github.com/ros2/examples/issues/20

Issue #20, opened 2015-04-29
Improve creation of service on Windows


wjwwood:
See: https://github.com/ros2/examples/pull/18

esteve:
This might be caused by http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#2132

PR #21, opened 2015-04-29
Remove static_cast

esteve:
Connects to #20 

@dirk-thomas @tfoote @wjwwood


dirk-thomas:
This seems to only revert the previous workaround from 53c425556152ceb2ef7c569ae6ba5ae0f944958c. Why does it work now?

Should we instead "git revert" that commit instead of having a new commit undoing the previous one?

esteve:
@dirk-thomas yes, it'll work after ros2/rclcpp#27 gets merged.

PR #22, opened 2015-04-30
Added example for the new sync parameters API

esteve:
Connects to ros2/ros2#28

@dirk-thomas @tfoote @wjwwood

### DIFF
--- a/userland/src/sync_parameters.cpp
+++ b/userland/src/sync_parameters.cpp
@@ -43,16 +43,20 @@ int main(int argc, char ** argv)
     std::cout << "Parameter value: ";
     switch (p.get_type()) {
       case rclcpp::parameter::ParameterType::PARAMETER_BOOL:
-        std::cout << "(bool): " << p.get_value<bool>() ? "true" : "false";
+        std::cout << "(bool): " <<
+        (p.get_value<rclcpp::parameter::ParameterType::PARAMETER_BOOL>() ? "true" : "false");
         break;
       case rclcpp::parameter::ParameterType::PARAMETER_INTEGER:
-        std::cout << "(integer): " << p.get_value<int64_t>();
+        std::cout << "(integer): " <<
+        p.get_value<rclcpp::parameter::ParameterType::PARAMETER_INTEGER>();
         break;
       case rclcpp::parameter::ParameterType::PARAMETER_DOUBLE:
-        std::cout << "(double): " << p.get_value<double>();
+        std::cout << "(double): " <<
+        p.get_value<rclcpp::parameter::ParameterType::PARAMETER_DOUBLE>();
         break;
       case rclcpp::parameter::ParameterType::PARAMETER_STRING:
-        std::cout << "(string): " << p.get_value<std::string>();
+        std::cout << "(string): " <<
+        p.get_value<rclcpp::parameter::ParameterType::PARAMETER_STRING>();
         break;
       case rclcpp::parameter::ParameterType::PARAMETER_BYTES:
         std::cout << "(bytes)";
--- a/userland/src/sync_parameters.cpp
+++ b/userland/src/sync_parameters.cpp
@@ -40,9 +40,9 @@ int main(int argc, char ** argv)
   auto sync_parameters_client = std::make_shared<rclcpp::parameter::SyncParametersClient>(node);
 
   auto parameter_variants = {
-    rclcpp::parameter::ParameterVariant("foo", int64_t(2)),
-    rclcpp::parameter::ParameterVariant("bar", std::string("hello")),
-    rclcpp::parameter::ParameterVariant("baz", double(1.45)),
+    rclcpp::parameter::ParameterVariant("foo", 2),
+    rclcpp::parameter::ParameterVariant("bar", "hello"),
+    rclcpp::parameter::ParameterVariant("baz", 1.45),
     rclcpp::parameter::ParameterVariant("foobar", true),
   };
 
--- a/userland/src/sync_parameters.cpp
+++ b/userland/src/sync_parameters.cpp
@@ -12,24 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <algorithm>
-#include <chrono>
-#include <cmath>
-#include <iostream>
-#include <string>
-#include <thread>
-
 #include <rclcpp/rclcpp.hpp>
-
-#include <simple_msgs/AllBuiltinTypes.h>
-#include <simple_msgs/AllDynamicArrayTypes.h>
-#include <simple_msgs/AllPrimitiveTypes.h>
-#include <simple_msgs/AllStaticArrayTypes.h>
-#include <simple_msgs/Nested.h>
-#include <simple_msgs/String.h>
-#include <simple_msgs/Uint32.h>
-
-#include <userland_msgs/AddTwoInts.h>
 
 int main(int argc, char ** argv)
 {
--- a/userland/src/sync_parameters.cpp
+++ b/userland/src/sync_parameters.cpp
@@ -1,0 +1,96 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


userland/src/sync_parameters.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::cout << "(bytes)";
+        break;
+      case rclcpp::parameter::ParameterType::PARAMETER_NOT_SET:
+      default:
+        std::cout << "INVALID";
+    }
+    std::cout << std::endl;
+  }
+
+  auto results3 = sync_parameters_client->list_parameters({{"foo", "bar"}}, 10);
```
wjwwood: Since we're using `auto`, it might be helpful to use more concrete names for the returned variable names. For instance, `results2` could be called just `parameters` or `parameter_variants` and `results3` could be called `parameters_and_prefixes` (to hint that it contains both). If you do that then `results1` can become just `results`. What do you think?

userland/src/sync_parameters.cpp
```diff
@@ -0,0 +1,96 @@
...
+    rclcpp::parameter::ParameterVariant("baz", double(1.45)),
+    rclcpp::parameter::ParameterVariant("foobar", true),
+  };
+
+  auto results1 = sync_parameters_client->set_parameters(parameters);
+  for (auto v : results1) {
+    if (!v.successful) {
+      std::cerr << v.reason << std::endl;
+    }
+  }
```
wjwwood: I think it would be good in the example to show how to associate the setting result and the parameter you requested to set, such that the message might be `Failed to set parameter 'foo' as a 'type' with value 'value': 'reason'`. The reason I think it's good to introduce this into the example is it shows users how they might accomplish this and if it turns out to be awkward for us to do it, then maybe we need to rethink what we return from that function or we need to provide a helper function to make it easier.

For instance, we might provide a set parameters result "packing" or "zipping" function, which would look like this:

```c++
auto results = sync_parameters_client->set_parameters(parameters);
for (auto & result : rclcpp::parameters::pack_set_parameter_results(parameters, results)) {
  if (!result.successful) {
    std::cerr << "Failed to set parameter '" << result.parameter.name
              << "' as a '" << result.parameter.type
              << "' with value '" << result.parameter.to_string()
              << "': '" << result.reason << "'" << std::endl;
  }
}
```

What do you guys think?

userland/src/sync_parameters.cpp
```diff
@@ -0,0 +1,96 @@
...
+  rclcpp::init(argc, argv);
+
+  auto node = rclcpp::Node::make_shared("parameters");
+
+  auto sync_parameters_client = std::make_shared<rclcpp::parameter::SyncParametersClient>(node);
+
+  auto parameters = {
+    rclcpp::parameter::ParameterVariant("foo", int64_t(2)),
+    rclcpp::parameter::ParameterVariant("bar", std::string("hello")),
+    rclcpp::parameter::ParameterVariant("baz", double(1.45)),
```
wjwwood: Are the `int64_t` and `double` calls required, or is it just to be explicit in the example? It might be worth trying to expose the user to the idea that a literal will be stored as an `int64_t` or a `double` implicitly. We might do this by just passing the literal, but tagging on a comment which says that it will be stored as a `double` or whatever. I'm not sure what the most instructive style would be for the user.

userland/src/sync_parameters.cpp
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  auto results1 = sync_parameters_client->set_parameters(parameters);
+  for (auto v : results1) {
+    if (!v.successful) {
+      std::cerr << v.reason << std::endl;
+    }
+  }
+
+  auto results2 = sync_parameters_client->get_parameters({{"foo", "baz"}});
```
wjwwood: Do you need two sets of `{}` here? I can never remember, initializer lists are still a bit confusing to me. Naively I would have thought that `{"foo", "bar"}` would be enough to initialize the vector of strings, but I remember struggling with it on VC++ and gcc/clang.

userland/src/sync_parameters.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  for (auto p : results2) {
```
wjwwood: Unless there is a reason to do a copy, I think it would be a good idea to use `auto & p`.

userland/src/sync_parameters.cpp
esteve: Sounds good. Done.

wjwwood:
Other than a few comments I think this looks really good. It's using the generated data structures where ever possible and seems quite consistent.

One other think that I noticed is that in this commit: https://github.com/ros2/rcl_interfaces/commit/9124b2149f9538c0637dfa652114b5099fdf5e48

You created a separate message for the list parameters result, could you elaborate on why that was needed?

After we iterate on the few comments I have +1.

userland/src/sync_parameters.cpp
esteve: This was a byproduct of ambiguous constructors in `ParameterVariant`, I added explicit constructors for all existing types and for `int` and `float`:

https://github.com/ros2/rclcpp/commit/7c54d94cbd3acf04768bcb2328d3855bebfa8344

userland/src/sync_parameters.cpp
esteve: Done.

userland/src/sync_parameters.cpp
esteve: I tend to use two sets of braces by default now, I remember having problems with just one set on gcc and we had the same problem with VC++. I'll try compiling this with only one set.

wjwwood: Yeah, I guess if `get_parameters` could take either a vector of strings or a string then `get_parameters({"foo"})` could be considered ambiguous. I think a similar scenario was the issue on Windows, but clang and gcc correctly (imo) say `get_parameters({"foo"})` matches the string version and `get_parameters({{"foo"}})` would be needed to match the vector of strings.

If we wanted to be explicit then we would want to do `get_parameters({{"foo"}, {"bar"}})`.

But since `get_parameters` doesn't yet take just a string, I think your latest commit is perfect.

esteve:
@tfoote and I thought that the API would look more symmetric, `list_parameters` was the only API call that returned results in more than one type, so we'd have to expose the `Response` object to the API, which IMO looks a bit awkward. But we can change it back if `ParameterListResult` is unnecessary.

wjwwood:
I'm actually fine with either, but if we keep the message, then I would say it should maybe be called `ListParametersResult`, so that it is clearer that it is associated with that service. That is unless we intend to use it in other places.

esteve:
@wjwwood We use `ParameterListResult` and `ParameterSetResult` instead because we didn't want them to look ambiguous with `ListParametersResponse` and `SetParametersResponse`

tfoote:
+1

userland/src/sync_parameters.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include <simple_msgs/AllBuiltinTypes.h>
+#include <simple_msgs/AllDynamicArrayTypes.h>
+#include <simple_msgs/AllPrimitiveTypes.h>
+#include <simple_msgs/AllStaticArrayTypes.h>
+#include <simple_msgs/Nested.h>
+#include <simple_msgs/String.h>
+#include <simple_msgs/Uint32.h>
+
+#include <userland_msgs/AddTwoInts.h>
```
dirk-thomas: Those includes look unnecessary.

dirk-thomas:
I would find two classes with the same name but e.g. in different namespaces less confusing then having `ListParametersResult` as well as `ParameterListResult`. The later version makes it less obvious to the user which one is used for what and why one is twisted.

userland/src/sync_parameters.cpp
esteve: Done.

dirk-thomas:
Beside the comment about the class names it looks good to me.

esteve:
I've renamed ParameterListResult and ParameterSetResult to ListParametersResult and SetParametersResult respectively in https://github.com/ros2/rcl_interfaces/pull/3

PR #26, opened 2015-05-26
add system_tests repo

dirk-thomas:
Connects to ros2/ros2#42

@esteve @tfoote @wjwwood Please review.


esteve:
+1, but I'm starting to think that the number of repositories is getting more complicated to manage.

tfoote:
+1

PR #28, opened 2015-05-29
refactor message generation

dirk-thomas:
Connects to ros2/ros2#48.

### DIFF
--- a/userland/src/add_two_ints_client.cpp
+++ b/userland/src/add_two_ints_client.cpp
@@ -21,15 +21,15 @@
 
 #include <rclcpp/rclcpp.hpp>
 
-#include <simple_msgs/AllBuiltinTypes.h>
-#include <simple_msgs/AllDynamicArrayTypes.h>
-#include <simple_msgs/AllPrimitiveTypes.h>
-#include <simple_msgs/AllStaticArrayTypes.h>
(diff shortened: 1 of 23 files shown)


esteve:
+1

userland/src/add_two_ints_client.cpp
```diff
@@ -37,8 +37,10 @@ int main(int argc, char ** argv)
 
   auto node = rclcpp::Node::make_shared("add_two_ints_client");
 
-  auto client = node->create_client<userland_msgs::AddTwoInts>("add_two_ints");
-  auto request = std::make_shared<userland_msgs::AddTwoInts::Request>();
+  auto client = node->create_client<
+userland_msgs::srv::AddTwoInts>("add_two_ints");
+  auto request = std::make_shared<
+userland_msgs::srv::AddTwoInts::Request>();
```
wjwwood: [style] Why are these lines lacking any leading whitespace? I would expect them to the be indented once from the indention level of the line they started on. (similar cases through out the pr)

dirk-thomas: Fixed 3716b39facc64c29b8a7577643d60b2c668c3ee6. Same for server part.

wjwwood:
+1

tfoote:
+1

PR #29, opened 2015-06-10
Subscribe to parameter events

esteve:
This PR adds an example of a subscription to a parameter event.

Connects to ros2/ros2#28

@dirk-thomas @jacquelinekay @tfoote @wjwwood

### DIFF
--- a/userland/src/async_parameters.cpp
+++ b/userland/src/async_parameters.cpp
@@ -54,15 +54,15 @@ int main(int argc, char ** argv)
         break;
       case rclcpp::parameter::ParameterType::PARAMETER_INTEGER:
         std::cout << "(integer): " <<
-        p.get_value<rclcpp::parameter::ParameterType::PARAMETER_INTEGER>();
+          p.get_value<rclcpp::parameter::ParameterType::PARAMETER_INTEGER>();
         break;
       case rclcpp::parameter::ParameterType::PARAMETER_DOUBLE:
         std::cout << "(double): " <<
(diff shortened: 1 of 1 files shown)


tfoote:
+1

wjwwood:
+1 This looks fine for now, but in the future we may want to hide this behind a `on_parameter_event(callback)` type function, so that we can do filtering for the user or to hide the implementation detail of the topic name used for this. On the other hand this is more dog food'y.

esteve:
@wjwwood agreed. I just added a new `on_parameter_event` method to `AsyncParameterClient` (https://github.com/ros2/rclcpp/commit/acdcf81a502179e2d4675640d2f322fd08dfcd61) that hides the creation of the subscription.

dirk-thomas:
+1

userland/src/async_parameters.cpp
```diff
@@ -84,6 +84,11 @@ int main(int argc, char ** argv)
     std::cout << "Parameter prefix: " << parameter_prefix << std::endl;
   }
 
+  auto callback = [](const std::shared_ptr<rcl_interfaces::msg::ParameterEvent> & event) -> void {
+      std::cout << "parameter foo updated" << std::endl;
```
tfoote: Should this print the name instead of foo?

esteve: Fixed. I ended up printing all the information from ParameterEvent (new, changed and deleted parameters)

PR #31, opened 2015-06-17
update message API

dirk-thomas:
@ros2/owners FYI

This PR implements the changes defined by ros2/design#33.


dirk-thomas:
I will go ahead and merge the changes since they only perform changes we agreed on in the article.

Issue #4, opened 2015-03-10
Refactor example and message packages


wjwwood:
We should split up the examples to be `rclc` or `rclcpp` specific.

AC:

- [x] Split `userland` into `rclc_examples` and `rclcpp_examples`
- [x] Refactor `simple_msgs`
 - Have a meeting to decide on the types here (perhaps get rid of `Int32` for example)
 - Make a `common_msgs` repository and move messages out of this `examples` repository.

wjwwood:
I think we should do this before moving on the the demos, and before we get a bunch of new outstanding branches. I think it will also speed up our build by removing some unused messages. So I've updated proposal:

New repository `common_interfaces` with packages:

**`builtin_interfaces`** (was `builtin_msgs`)
- Time.msg
- Duration.msg

**`std_interfaces`** (was `simple_msgs`)
- Drop these messages:
 - AllBuiltinTypes.msg
 - AllDynamicArrayTypes.msg
 - AllPrimitiveConstants.msg
 - AllPrimitiveTypes.msg
 - AllStaticArrayTypes.msg
 - Imu.msg (should probably be moved to a `sensor_interfaces`)
 - Imu32.msg
 - Intraprocess.msg
 - Nested.msg
 - Quaternion.msg (should probably be moved to a `geometry_interfaces`)
 - Quaternion32.msg
- Potentially drop these messages (depends on our decision):
  - Int32.msg
  - String.msg
  - Uint32.msg
  - Vector3Float.msg
  - Vector3Int.msg
- If these "base type messages" are not dropped, bring them up to parity with ROS 1's `std_msgs`.
- Add/keep messages from ROS 1's `std_msgs` and `std_srvs`:
  - ColorRGBA.msg
  - Empty.msg
  - Header.msg
  - MultiArrayDimension.msg
  - MultiArrayLayout.msg
  - Empty.srv
  - Trigger.srv

**`example_interfaces`** (was `userland_msgs`; could be moved to it's own repository, but not strictly necessary)
 - AddTwoInts.srv

Optionally we could "sync" more of the ROS 1 msg/src packages to the ROS 2 `common_interfaces` repository.

Changes to the `examples` repository:
- Remove the `builtin_msgs`, `simple_msgs`, and `userland_msgs` packages.
- Rename/split `userland` into `rclcpp_examples` and `rclc_examples` (`rclc` as needed).
- Remove unused or outdated examples like the intra-process example and the Imu examples.
[...]

PR #32, opened 2015-06-18
Refactor examples and interfaces

wjwwood:
I know the diff is hard to read in some places, but please just look at the resulting files as they are. I think it will be easiest to review it that way.

@esteve I also modified the services and parameters examples to be as concise as possible, I hope you don't mind. We can iterate on them if you like to make sure the API's are sufficiently represented. For instance, we might want to create additional parameter examples since I dropped the part of the example dealing with parameter events, since it was at the end and would never really print anything. It might even make sense to have an example just for listing parameters.

This also depends on https://github.com/ros2/rclcpp/pull/42.

Connects to ros2/examples#4


wjwwood:
@esteve I just pushed another commit that adds additional parameter examples to make up for the content I removed from the original pair.

dirk-thomas:
Does the `rclc_examples` not work anymore?

dirk-thomas:
Does the `ros1_like` Python code has enough value to keep it?

wjwwood:
> Does the rclc_examples not work anymore?

Before they were just source files; They weren't even listed in the `CMakeLists.txt`. But I will give the a try and fix the build if it is straight forward. I honestly don't remember what the issue was at this point.

> Does the ros1_like Python code has enough value to keep it?

I could create a disabled package like `rclc_examples` which holds them. I figured we could go back and get those examples when we needed them using the tag I created. What would everyone prefer?

esteve:
@wjwwood of course I don't mind, they look much better now! :-) Thanks for taking the time to make them more readable and concise.

+1

wjwwood:
I fixed up the `rclc_examples` in 640d426, but they still don't link because there are pieces of `rclc` which are not implemented yet. It's basically just an API, which is sufficient to build an example on, but won't link and run. So I left the `AMENT_IGNORE` for now.

wjwwood:
I still need to test out the new examples, to make sure they work as expected, I'm doing that now.

wjwwood:
Ok, the pub/sub and service examples work. There are some issues with the parameter examples, which @esteve and I looked at for a while today, but I think they'll take some more work to figure out and the issues are with the rclcpp implementation (not caused by the demo). So I'd like to go ahead and merge this refactoring without fixing them for now.

PR #35, opened 2015-06-25
Sync parameter events callbacks

esteve:
Adds examples for `on_parameter_event`

Connects to ros2/rclcpp#49


rclcpp_examples/src/parameters/set_and_get_parameters.cpp
```diff
@@ -27,7 +27,7 @@ int main(int argc, char ** argv)
 
   auto parameters_client = std::make_shared<rclcpp::parameter_client::SyncParametersClient>(node);
 
-  // Set several differnet types of parameters.
+  // Set several different types of parameters.
```
wjwwood: Doh.

wjwwood:
lgtm

rclcpp_examples/src/parameters/set_and_get_parameters.cpp
esteve: I could be wrong, though. Maybe there's the internet, the samenet and the differnet :stuck_out_tongue_winking_eye:

tfoote:
+1

dirk-thomas:
lgtm

jacquelinekay:
looks fine. At some point we should consider adding a lot more comments to all the examples, maybe giving a summary of what the example does at the top and a little bit of motivation about why it is exists. But this could be done at a later point, since the selection of examples is still evolving.

PR #37, opened 2015-07-17
Example for switching MemoryStrategy

jacquelinekay:
Needs https://github.com/ros2/rclcpp/pull/60

### DIFF
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -26,7 +26,7 @@ add_executable_for_each_rmw_impl(talker src/topics/talker.cpp)
 add_executable_for_each_rmw_impl(listener src/topics/listener.cpp)
 
 ## Example of subscribe with topics and customized executor
-add_executable_for_each_rmw_impl(listener_exec src/topics/listener_exec.cpp)
+add_executable_for_each_rmw_impl(listener_memory src/topics/listener_memory.cpp)
 
 ##
 ## Examples of Request/Response with Services
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -25,6 +25,9 @@ include(cmake/add_executable_for_each_rmw_impl.cmake)
 add_executable_for_each_rmw_impl(talker src/topics/talker.cpp)
 add_executable_for_each_rmw_impl(listener src/topics/listener.cpp)
 
+## Example of subscribe with topics and customized executor
+add_executable_for_each_rmw_impl(listener_exec src/topics/listener_exec.cpp)
+
 ##
 ## Examples of Request/Response with Services
 ##
--- a/rclcpp_examples/src/topics/listener_exec.cpp
+++ b/rclcpp_examples/src/topics/listener_exec.cpp
@@ -25,14 +25,13 @@ using rclcpp::memory_strategies::static_memory_strategy::StaticMemoryStrategy;
 
 void chatterCallback(const std_interfaces::msg::String::ConstSharedPtr & msg)
 {
-  std::cout << "I heard: [" << msg->data << "]" << std::endl << std::flush;
+  printf("I heard: [%s]\n", msg->data.c_str());
 }
 
 int main(int argc, char * argv[])
 {
   rclcpp::init(argc, argv);
-  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy =
-    rclcpp::memory_strategy::create_default_strategy();
+  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy = nullptr;
   if (argc > 1) {
     std::string argument(argv[1]);
     if (argument == "static") {
@@ -44,6 +43,9 @@ int main(int argc, char * argv[])
       std::cout << "Warning: unknown argument. " << std::endl;
       std::cout << "Setting memory allocation strategy to default (dynamic)." << std::endl;
     }
+  }
+  if (memory_strategy == nullptr) {
+    memory_strategy = rclcpp::memory_strategy::create_default_strategy();
   }
 
   auto node = rclcpp::Node::make_shared("listener_exec");
--- a/rclcpp_examples/src/topics/listener_exec.cpp
+++ b/rclcpp_examples/src/topics/listener_exec.cpp
@@ -22,7 +22,6 @@
 #include <std_interfaces/msg/string.hpp>
 
 using rclcpp::memory_strategies::static_memory_strategy::StaticMemoryStrategy;
-using rclcpp::memory_strategies::dynamic_memory_strategy::DynamicMemoryStrategy;
 
 void chatterCallback(const std_interfaces::msg::String::ConstSharedPtr & msg)
 {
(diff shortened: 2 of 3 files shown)


rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,65 @@
...
+  std::cout << "I heard: [" << msg->data << "]" << std::endl;
+}
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy =
+      rclcpp::memory_strategy::create_default_strategy();
+  if (argc > 1)
+  {
```
dirk-thomas: Style. Same below.

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,65 @@
...
+    std::string argument(argv[1]);
+    if (argument == "static")
+    {
+      std::cout << "Setting memory allocation strategy to 'static'." << std::endl;
+      memory_strategy = std::make_shared<StaticMemoryStrategy>(StaticMemoryStrategy());
```
dirk-thomas: Is it possible to select the static strategy without creating a default strategy before?

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+  rclcpp::init(argc, argv);
+  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy =
+      rclcpp::memory_strategy::create_default_strategy();
+  if (argc > 1)
+  {
+    std::string argument(argv[1]);
+    if (argument == "static")
+    {
+      std::cout << "Setting memory allocation strategy to 'static'." << std::endl;
```
dirk-thomas: Please use `printf` instead to avoid interleaving.

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,65 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
```
dirk-thomas: 2015

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,65 @@
...
+    else
+    {
+      std::cout << "Warning: unknown argument. " << std::endl;
+      std::cout << "Setting memory allocation strategy to default (dynamic)." << std::endl;
+    }
+  }
+
+  auto node = rclcpp::Node::make_shared("listener");
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.set_memory_strategy(memory_strategy);
```
wjwwood: I know I may have originally suggested that we have a function to set the memory strategy, but now I'm wondinger if this should be a parameter to the constructor and prevent it from being changed after creation of the executor? I guess it has to do with the scope of the memory strategy and whether or not it affects the behavior of the constructor, destructor, add_node, etc... Currently I know it only affects the spin methods, so maybe this is solved with documentation about what the memory strategy can influence.

rclcpp_examples/src/topics/listener_exec.cpp
wjwwood: -0, there shouldn't be any interleaved output here and std::cout is more idiomatic C++. A better solution, since its a single string, would be to just do this:

```c++
std::cout << "Setting memory allocation strategy to 'static'.\n" << std::flush;
```

rclcpp_examples/src/topics/listener_exec.cpp
wjwwood: I'd recommend setting it to `nullptr` on line 34 and then creating a default memory strategy if it is still `nullptr` on line ~54.

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,65 @@
...
+      std::cout << "Setting memory allocation strategy to 'dynamic'." << std::endl;
+    }
+    else
+    {
+      std::cout << "Warning: unknown argument. " << std::endl;
+      std::cout << "Setting memory allocation strategy to default (dynamic)." << std::endl;
+    }
+  }
+
+  auto node = rclcpp::Node::make_shared("listener");
```
wjwwood: For consistency should this be `listener_exec`?

wjwwood:
+1, aside from comments.

rclcpp_examples/src/topics/listener_exec.cpp
dirk-thomas: This listener might be used together with a publisher and then result in the mentioned problem. Therefore we should consider to always use a safe way to output messages.

rclcpp_examples/src/topics/listener_exec.cpp
jacquelinekay: Yeah, I initially went back and forth on this: at first I only set the memory strategy in the constructor, then I added the `set_memory_strategy` function. Changing the memory strategy while the executor is running seems like a bad idea in most cases, but I'll wait off on removing this function from the executor until the `MemoryStrategy`

jacquelinekay: OK, in order to do this I had to change the constructor of `SingleThreadedExecutor` to pass a memory strategy to the parent constructor in https://github.com/ros2/rclcpp/pull/60. I also added the argument to the constructor of `MultiThreadedExecutor`, for good measure.

rclcpp_examples/CMakeLists.txt
```diff
@@ -25,6 +25,9 @@ include(cmake/add_executable_for_each_rmw_impl.cmake)
 add_executable_for_each_rmw_impl(talker src/topics/talker.cpp)
 add_executable_for_each_rmw_impl(listener src/topics/listener.cpp)
 
+## Example of subscribe with topics and customized executor
+add_executable_for_each_rmw_impl(listener_exec src/topics/listener_exec.cpp)
```
dirk-thomas: Why is the node called `listener_exec`? Since it is different from the "normal" listener by using different memory strategies I would propose calling it `listener_memory`.

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,60 @@
...
+  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy = nullptr;
+  if (argc > 1) {
+    std::string argument(argv[1]);
+    if (argument == "static") {
+      std::cout << "Setting memory allocation strategy to 'static'." << std::endl;
+      memory_strategy = std::make_shared<StaticMemoryStrategy>(StaticMemoryStrategy());
+    } else if (argument == "dynamic") {
+      std::cout << "Setting memory allocation strategy to 'dynamic'." << std::endl;
+    } else {
+      std::cout << "Warning: unknown argument. " << std::endl;
```
dirk-thomas: These lines should also use `printf`.

This one should even go to `stderr`.

rclcpp_examples/src/topics/listener_exec.cpp
```diff
@@ -0,0 +1,60 @@
...
+      std::cout << "Setting memory allocation strategy to default (dynamic)." << std::endl;
+    }
+  }
+  if (memory_strategy == nullptr) {
```
dirk-thomas: Should this use `!memory_strategy` like in other places?

dirk-thomas:
+1 aside from comments.

PR #39, opened 2015-07-22
Message memory strategy example

jacquelinekay:
Needs https://github.com/ros2/rclcpp/pull/64


rclcpp_examples/src/topics/listener_memory.cpp
```diff
@@ -23,9 +23,11 @@
 
 using rclcpp::memory_strategies::static_memory_strategy::StaticMemoryStrategy;
 
-void chatterCallback(const std_interfaces::msg::String::ConstSharedPtr & msg)
+unsigned int messages_received = 0;
```
wjwwood: I would recommend using `size_t` here.

rclcpp_examples/src/topics/listener_single_msg.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <std_interfaces/msg/string.hpp>
+
+using namespace rclcpp::single_msg_memory_strategy;
+using rclcpp::memory_strategies::static_memory_strategy::StaticMemoryStrategy;
+
+unsigned int messages_received = 0;
+
+void chatterCallback(const std_interfaces::msg::String::ConstSharedPtr & /*msg*/)
+{
+  ++messages_received;
```
wjwwood: Would it make sense to print out the address in msg to show that the same message is being used all the time?

rclcpp_examples/src/topics/listener_memory.cpp
```diff
@@ -23,9 +23,11 @@
...
+
+void chatterCallback(const std_interfaces::msg::String::ConstSharedPtr & /*msg*/)
 {
-  printf("I heard: [%s]\n", msg->data.c_str());
+  ++messages_received;
```
wjwwood: Just out of curiosity, why aren't you printing out what you've received here?

rclcpp_examples/src/topics/listener_memory.cpp
```diff
@@ -56,5 +58,7 @@ int main(int argc, char * argv[])
 
   executor.spin();
 
+  printf("Received %d messages.\n", messages_received);
```
esteve: You can also use `std::to_string` with `std::cout` instead of `printf` to make it more C++ idiomatic, but it's just a cosmetic change.

rclcpp_examples/src/topics/listener_single_msg.cpp
esteve: +1 sounds like a good idea.

rclcpp_examples/src/topics/listener_single_msg.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy = nullptr;
+  if (argc > 1) {
+    std::string argument(argv[1]);
+    if (argument == "static") {
+      printf("Setting memory allocation strategy to 'static'.\n");
```
esteve: Just curiosity, are you using `printf` to avoid the extra allocations that come with `std::cout` and `std::cerr` ?

esteve:
+1

rclcpp_examples/src/topics/listener_memory.cpp
jacquelinekay: I'm working towards making this example parallel to the examples I used for real-time testing, which avoid screen I/O because of the potential for nondeterministic blocking.

rclcpp_examples/src/topics/listener_single_msg.cpp
jacquelinekay: I'm using `printf` because @dirk-thomas always comments that it should be used instead of `std::cout` to avoid interleaving. I'm not concerned about allocations here because it's not during the execution loop of the node.

rclcpp_examples/src/topics/listener_single_msg.cpp
jacquelinekay: Yeah, I do like this idea, I think it makes sense for this example.

rclcpp_examples/src/topics/listener_single_msg.cpp
esteve: That's right, every call to `<<` may be interleaved if multiple threads use `std::cout`. Interleaving can be avoided using a `std::stringstream`, though, but it's not worth the hassle. Thanks for the explanation.

rclcpp_examples/src/topics/listener_memory.cpp
esteve: Disregard this comment, I see why you're using `printf`

rclcpp_examples/src/topics/listener_single_msg.cpp
jacquelinekay: well, I made this change (haven't pushed it yet) and it turns out the address of the message is not the same every time. I'll do some more investigating on the branch of rclcpp to see why that is.

jacquelinekay: ok, the problem was that the base class implementation of `borrow_message` was being called, which was allocating a new message. Turns out that I was slicing the `SingleMessageMemoryStrategy` object before passing it to the node. (https://stackoverflow.com/questions/274626/what-is-object-slicing)

This line fixes it:
https://github.com/ros2/examples/commit/23efc24ab542deb17034fecc28ff8222388e1375#diff-2ceecd0a60d163175f8523de782353ffR64

You should now be able to run the examples and see the same pointer address printed out in the callback of `listener_single_msg`.

rclcpp_examples/src/topics/listener_single_msg.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  rclcpp::executors::SingleThreadedExecutor executor(memory_strategy);
+  auto node = rclcpp::Node::make_shared("listener_single_msg");
+
+  executor.add_node(node);
+
+  SingleMessageMemoryStrategy<std_interfaces::msg::String> msg_strategy;
+  auto msg_strategy_ptr =
+    std::make_shared<SingleMessageMemoryStrategy<std_interfaces::msg::String>>(
+    msg_strategy);
```
wjwwood: These two statements could be simplified as:

```c++
auto msg_strategy_ptr = std::make_shared<SingleMessageMemoryStrategy<std_interfaces::msg::String>>();
```

or

```c++
auto msg_strategy_ptr = SingleMessageMemoryStrategy<std_interfaces::msg::String>::SharedPtr(
  new SingleMessageMemoryStrategy<std_interfaces::msg::String>());
```

The current pair of statements are sort of wasteful since it expands to essentially this:

```c++
SingleMessageMemoryStrategy<std_interfaces::msg::String> msg_strategy;
SingleMessageMemoryStrategy<std_interfaces::msg::String>::SharedPtr msg_strategy_ptr(new SingleMessageMemoryStrategy<std_interfaces::msg::String>());
*msg_strategy_ptr = msg_strategy;  // implicit copy
```

rclcpp_examples/src/topics/listener_single_msg.cpp
wjwwood: > You should now be able to run the examples and see the same pointer address printed out in the callback of listener_single_msg.

Neat :smile:, glad I mentioned it!

jacquelinekay:
I refactored the examples a bit to be in line with the changes in ros2/rclcpp#64. Most notably, I've added a `LargeFixed` message type that is used, instead of the string example, since the `MessagePoolMemoryStrategy` does not allow non-fixed message types. weirdly enough, I see the same memory address being printed in both examples now (meaning that the default memory strategy is also only using one address).

jacquelinekay:
@wjwwood, if these examples look OK to you, I'll merge it by the end of today.

wjwwood:
lgtm, +1

dirk-thomas:
@jacquelinekay With the windows build being back up again please revisit the new warning introduced by this PR.

PR #45, opened 2015-08-10
Change print format for size_t

jacquelinekay:
http://ci.ros2.org/job/ros2_batch_ci_windows/136/

8 warnings are fixed:

http://ci.ros2.org/job/ros2_batch_ci_windows/136/warnings34Result/fixed/


dirk-thomas:
+1

PR #47, opened 2015-08-13
example of method chaining for object pool bounds struct

jacquelinekay:
requires ros2/rclcpp/pull/82


dirk-thomas:
+1

wjwwood:
+1

Did you run uncrustify over this? I find the line wrapping of the chained methods strange looking, but I don't know what else you might do.

jacquelinekay:
Yep, I tried indenting four spaces on the second line and uncrustify
corrected it.

jacquelinekay:
I got uncrustify to accept an indented second line. I think it looks better now.

PR #51, opened 2015-08-23
add gmock_vendor repo

dirk-thomas:


dirk-thomas:
Replaced by ros2/ros2#105.

PR #54, opened 2015-08-26
add ros1_bridge repo

dirk-thomas:
Depends on ros2/ros1_bridge#8 to only build if ROS 1 is sourced.


tfoote:
+1

Do we need to add the ros1 packages to our bootstrap now?

dirk-thomas:
If the user does not want to build the bridge: no.
If the user does want to build the bridge: yes.

tfoote:
My question was for our default builds, and we decided no.

PR #59, opened 2015-09-09
Fix CTRL-C in client examples

jacquelinekay:
Connects to ros2/rmw_connext#97

The global interrupt guard condition is working for clients/spin_until_future_complete, but the problem was in how the resulting future object was used. If you call `get` on a future that isn't complete, the thread will continue blocking on that future, even though the CTRL-C interrupt was already triggered. This pull request checks the status of the future before accessing it, to avoid blocking.

It may be wise to refactor the `spin_until_future_complete` call to return a more informative error code if the call was interrupted. I see this pull request as more of a workaround to a bigger issue in the rclcpp API.

### DIFF
--- a/rclcpp_examples/src/services/add_two_ints_client.cpp
+++ b/rclcpp_examples/src/services/add_two_ints_client.cpp
@@ -25,7 +25,11 @@ example_interfaces::srv::AddTwoInts_Response::SharedPtr send_request(
   example_interfaces::srv::AddTwoInts_Request::SharedPtr request)
 {
   auto result = client->async_send_request(request);
-  return rclcpp::spin_until_future_complete(node, result).get();
+  rclcpp::spin_until_future_complete(node, result);
+  if (result.wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
+    return result.get();
+  }
+  return NULL;
 }
 
 int main(int argc, char ** argv)
@@ -41,7 +45,12 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): make it like `client->send_request(node, request)->sum`
   // TODO(wjwwood): consider error condition
-  std::cout << "Result of add_two_ints: " << send_request(node, client, request)->sum << std::endl;
+  auto result = send_request(node, client, request);
+  if (result) {
+    printf("Result of add_two_ints: %d\n", result->sum);
+  } else {
+    printf("add_two_ints_client was interrupted. Exiting.\n");
+  }
 
   return 0;
 }
--- a/rclcpp_examples/src/services/add_two_ints_client_async.cpp
+++ b/rclcpp_examples/src/services/add_two_ints_client_async.cpp
@@ -33,7 +33,11 @@ int main(int argc, char ** argv)
 
   rclcpp::spin_until_future_complete(node, result);  // Wait for the result.
 
-  std::cout << "Result of add_two_ints: " << result.get()->sum << std::endl;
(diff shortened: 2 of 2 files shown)


wjwwood:
I'm testing this out now.

wjwwood:
Sorry, I've run into a new problem (after updating all my packages) where I cannot generate more than one version of the binaries (I'm missing `*__rmw_*`). And it seems that connext is not finding the license file anymore. I'm still trying to test this pr asap.

dirk-thomas:
I just tried it with Connext and I was able to Ctrl-C the client and get the message:

    add_two_ints_client was interrupted. Exiting.

The server exited as well (without a message) after Ctrl-C.

But their seems to be no service invocation.
I don't get any output beside the message above (and the RTI license information).

rclcpp_examples/src/services/add_two_ints_client.cpp
```diff
@@ -41,7 +45,12 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): make it like `client->send_request(node, request)->sum`
   // TODO(wjwwood): consider error condition
-  std::cout << "Result of add_two_ints: " << send_request(node, client, request)->sum << std::endl;
+  auto result = send_request(node, client, request);
+  if (result) {
+    std::cout << "Result of add_two_ints: " << result->sum << std::endl;
```
dirk-thomas: New / updated code should use `printf` instead.

dirk-thomas:
Just to clarify: even without the patch the service invocation doesn't happen for me with Connext (static).

jacquelinekay:
Yes, this PR fixes ros2/rmw_connext#97 but it does not fix any of the problems with service invocations in Connext. I found that the add_two_ints example service only works if I start the server and wait a few seconds before starting the client. I also made an working example where the service and client are in the same node (but this is not a very useful configuration).

I'm debugging some of the problems with services right now, but the fix isn't obvious to me yet.

wjwwood:
I'm able to get the service call to complete with all three major vendors.

wjwwood:
And I'm able to ctrl-c the client when the server is not running.

@dirk-thomas I did notice that if you start the client first, then the server, then it will not complete.

wjwwood:
I didn't try the behavior about ctrl-c before the patch, but +1.

dirk-thomas:
With a few seconds wait between starting the server and client it works for me too. Just being to impatient :wink: +1

rclcpp_examples/src/services/add_two_ints_client.cpp
```diff
@@ -41,7 +45,12 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): make it like `client->send_request(node, request)->sum`
   // TODO(wjwwood): consider error condition
-  std::cout << "Result of add_two_ints: " << send_request(node, client, request)->sum << std::endl;
+  auto result = send_request(node, client, request);
+  if (result) {
+    printf("Result of add_two_ints: %d\n", result->sum);
```
dirk-thomas: Depending on what type `sum` is this might result in a compiler warning on Windows...

jacquelinekay: Ugh. The message specification for the service says that `sum` is an int64. Aren't the warnings usually for unsigned types?

I guess there's one way to find out for sure... http://ci.ros2.org/job/ros2_batch_ci_windows/383/

PR #61, opened 2015-09-09
fix warnings

jacquelinekay:
I introduced new warnings in ros2/examples/pull/59. This is what happens when you don't launch CI...

http://ci.ros2.org/job/ros2_batch_ci_windows/384/

I'm in favor of squashing this into the commit that introduced the warning.

### DIFF
--- a/rclcpp_examples/src/services/add_two_ints_client.cpp
+++ b/rclcpp_examples/src/services/add_two_ints_client.cpp
@@ -47,7 +47,7 @@ int main(int argc, char ** argv)
   // TODO(wjwwood): consider error condition
   auto result = send_request(node, client, request);
   if (result) {
-    printf("Result of add_two_ints: %d\n", result->sum);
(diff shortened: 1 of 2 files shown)


tfoote:
+1

dirk-thomas:
It seems to still result in warnings on OS X: http://ci.ros2.org/job/ros2_batch_ci_osx_nightly/39/warnings7Result/new/

dirk-thomas:
Is this (#61) as well as #59 now covered in the commit 10162bf0f69faea613045c46c84a7299fa7dd818 on the master branch?

If yes, I assume you squashed commits which had already been on the master branch? If that is the case: we should *not* force push to master and that's what you had to do in order to squash the already pushed separate commits.

jacquelinekay:
Yes, I squashed the last three merged pull requests (#62, #61, #59) into
one commit by force pushing to master. Next time I will avoid force pushing
to master and instead make the changes "atomically" in one PR.

PR #67, opened 2015-10-27
remove pool memory strategy examples

jacquelinekay:
I will consider adding them back when there's an acceptable replacement that uses the new MemoryStrategy API.

Connects to ros2/rclcpp#137


esteve:
+1

PR #68, opened 2015-11-03
Remove extra braces. Wait for the correct future.

esteve:
This fixes two issues in the example.

### DIFF
--- a/rclcpp_examples/src/parameters/set_and_get_parameters_async.cpp
+++ b/rclcpp_examples/src/parameters/set_and_get_parameters_async.cpp
@@ -49,8 +49,8 @@ int main(int argc, char ** argv)
   }
 
   // Get a few of the parameters just set.
-  auto parameters = parameters_client->get_parameters({{"foo", "baz"}});
-  if (rclcpp::spin_until_future_complete(node, results) !=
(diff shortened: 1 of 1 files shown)


tfoote:
+1

wjwwood:
Lgtm

rclcpp_examples/src/parameters/set_and_get_parameters_async.cpp
```diff
@@ -49,8 +49,8 @@ int main(int argc, char ** argv)
   }
 
   // Get a few of the parameters just set.
-  auto parameters = parameters_client->get_parameters({{"foo", "baz"}});
-  if (rclcpp::spin_until_future_complete(node, results) !=
+  auto parameters = parameters_client->get_parameters({"foo", "baz"});
```
dirk-thomas: Is this change necessary to fix anything?

Weren't the extra braces necessary for Windows? We also do this in various other locations: https://github.com/ros2/system_tests/blob/master/test_communication/test/message_fixtures.hpp#L130

esteve: Yes, otherwise the example fails with an invalid iterator range error.

PR #69, opened 2015-11-03
Ensure that the on_parameter_event subscription is triggered

esteve:
This should report all the parameter events.

### DIFF
--- a/rclcpp_examples/src/parameters/parameter_events.cpp
+++ b/rclcpp_examples/src/parameters/parameter_events.cpp
@@ -63,5 +63,9 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): Create and use delete_parameter
 
+  rclcpp::sleep_for(100_ms);
+
+  rclcpp::spin_some(node);
+
   return 0;
 }
--- a/rclcpp_examples/src/parameters/parameter_events_async.cpp
+++ b/rclcpp_examples/src/parameters/parameter_events_async.cpp
@@ -65,7 +65,9 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): Create and use delete_parameter
 
-  rclcpp::spin(node);
+  rclcpp::sleep_for(100_ms);
+
+  rclcpp::spin_some(node);
 
   return 0;
 }


esteve:
This will spin until the user presses Ctrl-C, I don't know if we should make it more user-friendly and exit once the `on_parameter_event` callback receives all the events.

jacquelinekay:
I agree that the program should exit once the callback receives all events.

wjwwood:
What about something like this:

```diff
diff --git a/rclcpp_examples/src/parameters/parameter_events.cpp b/rclcpp_examples/src/parameters/parameter_events.cpp
index c28adcf..b1c2306 100644
--- a/rclcpp_examples/src/parameters/parameter_events.cpp
+++ b/rclcpp_examples/src/parameters/parameter_events.cpp
@@ -22,6 +22,9 @@ void on_parameter_event(const rcl_interfaces::msg::ParameterEvent::SharedPtr eve
   std::cout << "Parameter event:" << std::endl << " new parameters:" << std::endl;
   for (auto & new_parameter : event->new_parameters) {
     std::cout << "  " << new_parameter.name << std::endl;
+    if (new_parameter.name == "shutdown" && new_parameter.value) {
+      rclcpp::shutdown();
+    }
   }
   std::cout << " changed parameters:" << std::endl;
   for (auto & changed_parameter : event->changed_parameters) {
@@ -59,6 +62,7 @@ int main(int argc, char ** argv)
   set_parameters_results = parameters_client->set_parameters({
     rclcpp::parameter::ParameterVariant("foo", 3),
     rclcpp::parameter::ParameterVariant("bar", "world"),
+    rclcpp::parameter::ParameterVariant("shutdown", true),
   });
 
   // TODO(wjwwood): Create and use delete_parameter

```

wjwwood:
I edited my previous post.

esteve:
@wjwwood looks good, will apply your code to `parameter_events` and `parameter_events_async`

esteve:
Unfortunately, the `new_parameters` event comes with all at once, so the callback will call `rclcpp::shutdown` too early. I'm working on a different approach.

wjwwood:
Just set a bool `true` when the "shutdown" param is set, and shutdown at the end of the callback if the bool is `true`?

esteve:
I've updated the callback to check that all the parameters have been received before exiting. Depends on https://github.com/ros2/rclcpp/pull/143

rclcpp_examples/src/parameters/parameter_events.cpp
```diff
@@ -16,20 +16,36 @@
...
+  const std::set<std::string> & expected_deleted_parameters,
+  std::set<std::string> & received_new_parameters,
+  std::set<std::string> & received_changed_parameters,
+  std::set<std::string> & received_deleted_parameters,
+  const rcl_interfaces::msg::ParameterEvent::SharedPtr event)
 {
   // TODO(wjwwood): The message should have an operator<<, which would replace all of this.
   std::cout << "Parameter event:" << std::endl << " new parameters:" << std::endl;
   for (auto & new_parameter : event->new_parameters) {
     std::cout << "  " << new_parameter.name << std::endl;
```
dirk-thomas: Unrelated to the PR itself: `printf`

Same below.

rclcpp_examples/src/parameters/parameter_events_async.cpp
```diff
@@ -16,20 +16,36 @@
...
   }
   std::cout << " deleted parameters:" << std::endl;
   for (auto & deleted_parameter : event->deleted_parameters) {
     std::cout << "  " << deleted_parameter.name << std::endl;
+    received_deleted_parameters.insert(deleted_parameter.name);
+  }
+
+  if (expected_new_parameters == received_new_parameters &&
+    expected_changed_parameters == received_changed_parameters &&
+    expected_deleted_parameters == received_deleted_parameters) {
```
dirk-thomas: Please avoid this kind of indentation since there is no difference between the condition and the following block.

esteve: What would be the correct indentation?

dirk-thomas: One option would be:

    if (
      cond &&
      cond
    ) {
      ...
    }

esteve: Done and squashed.

wjwwood: I'd do this:

```c++
if (expected_new_parameters == received_new_parameters &&
    expected_changed_parameters == received_changed_parameters &&
    expected_deleted_parameters == received_deleted_parameters)
{
  // ...
}
```

Or this:

```c++
if (
  expected_new_parameters == received_new_parameters &&
  expected_changed_parameters == received_changed_parameters &&
  expected_deleted_parameters == received_deleted_parameters)
{
  // ...
}
```

But frankly multiline conditionals within if statements are just ugly (in C++ and Python).

esteve: I've updated the code to follow the second style.

wjwwood:
I think the result of this pr makes the example too complicated. It's meant to be a simple and strait forward as possible, and to that end I think we should avoid sets and lambda's. Also if for some reason one of the parameters doesn't arrive it will never exit spin.

Why not just call shutdown when a final parameter is set?

One other thought, did you try just running `rclcpp::spin_some(node)` at the end?

esteve:
`rclcpp::spin_some` seems to suffice. Updated and squashed.

rclcpp_examples/src/parameters/parameter_events.cpp
```diff
@@ -63,5 +63,7 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): Create and use delete_parameter
 
+  rclcpp::spin_some(node);
```
dirk-thomas: I expect this to be a race condition. Not all events might be available yet so potentially only some are handled here. I would suggest to add a sleep before, e.g. 100ms.

esteve: Done.

wjwwood:
lgtm

dirk-thomas:
lgtm

PR #71, opened 2015-11-05
toggle intra-process in allocator example

jacquelinekay:
For the problem fixed by ros2/system_tests#65, OSX failed to compile because somewhere the compiler was trying to specialize custom allocator `construct` for a const pointer, which led to a placement new into a const pointer, which is not allowed. The fix was relatively straightforward, but I wondered why OSX could compile the allocator example, especially because the allocator in that example has no construct or destroy methods. This is because the intra-process codepath was commented out in the allocator example. IntraProcessMangerState passes the custom allocator to a few standard library structures (such as map) which were the origin of the bug fixed by the system_tests pull requests.

Therefore this pull request does two things:

- Add construct and destroy using the same enable_if pattern used in ros2/system_tests#65
- Toggle intra-process with a command line argument instead of telling the user to comment it out, so that codepath gets compiled every time.

### DIFF
--- a/rclcpp_examples/src/topics/allocator_example.cpp
+++ b/rclcpp_examples/src/topics/allocator_example.cpp
@@ -74,6 +74,21 @@ public:
     std::free(ptr);
   }
 
+  template<typename U, typename ... Args,
+  typename std::enable_if<!std::is_const<U>::value>::type * = nullptr>
+  void
+  construct(U * ptr, Args && ... args)
+  {
+    ::new(ptr)U(std::forward<Args>(args) ...);
+  }
+
+  template<typename U>
+  void
+  destroy(U * ptr)
+  {
+    ptr->~U();
+  }
+
   template<typename U>
   struct rebind
   {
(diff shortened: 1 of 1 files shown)


esteve:
+1

rclcpp_examples/src/topics/allocator_example.cpp
```diff
@@ -138,16 +153,18 @@ int main(int argc, char ** argv)
...
-  auto context = rclcpp::contexts::default_context::get_global_default_context();
-  auto ipm_state =
-    std::make_shared<rclcpp::intra_process_manager::IntraProcessManagerState<MyAllocator<>>>();
-  // Constructs the intra-process manager with a custom allocator.
-  context->get_sub_context<rclcpp::intra_process_manager::IntraProcessManager>(ipm_state);
-  auto node = rclcpp::Node::make_shared("allocator_example", true);
-  */
+  rclcpp::Node::SharedPtr node;
+
+  if (argc > 1 && strstr(argv[1], "intra") != NULL) {
```
dirk-thomas: What is `strstr(argv[1], "intra") != NULL` checking for? Can't this be more specific?

jacquelinekay: lazy C-style substring. I'll change it to something cleaner.

jacquelinekay: I've changed it to accept "intra", "intra_process", "intra-process" and "intraprocess" as a single positional argument that turns the intra-process pipeline on. It also prints the mode that was chosen.

jacquelinekay:
Jenkins:

http://ci.ros2.org/job/ros2_batch_ci_linux/567/
http://ci.ros2.org/job/ros2_batch_ci_osx/441/

The Windows build is broken on nightly and the spin_timeout tests probably regressed because of ros2/rclcpp#146.

PR #79, opened 2015-12-07
add missing dependency on rmw_implementation_cmake

dirk-thomas:


dirk-thomas:
Since I need this for moving forward with the type support refactoring I will go ahead and merge it.

Feel free to comment afterwards.

PR #82, opened 2015-12-11
subscribing to bound method in example

tfoote:
Add an example of subscribing with a bind to a class method. 

This not working. It is an example for 

Connects to ros2/rclcpp#173

### DIFF
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -34,6 +34,7 @@ macro(targets)
   custom_executable(topics talker)
   custom_executable(topics listener)
   custom_executable(topics listener_best_effort)
+  custom_executable(topics listener_class)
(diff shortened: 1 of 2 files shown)


esteve:
@tfoote I've rebased this branch to fix the conflicts. Let me know if you're ok with that.

jacquelinekay:
I'm not sure if we need this example, since std::bind is now covered by the test you wrote and I don't want to add unnecessary code to maintain. Maybe we could make a note of it somewhere in the migration guide that std::bind is supported but lambdas are usually preferred as better C++11 practice, since most ROS users are familiar with this pattern.

tfoote:
We can close  this in deference to https://github.com/ros2/system_tests/issues/88 which is a unit test. It is valuable to show how to use a member based callback. But I think that will be more important when we get to end user documentation rather than our in progress examples.

PR #87, opened 2016-01-21
Added examples that exercise the rclpy API

esteve:
Added examples that exercise the rclpy API.

### DIFF
--- a/rclpy_examples/talker_py.py
+++ b/rclpy_examples/talker_py.py
@@ -1,0 +1,44 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from time import sleep
+
+import rclpy
+from rclpy.qos import qos_profile_default
+from std_msgs.msg import String
+
+
+def main(args=None):
+    if args is None:
+        args = sys.argv
+
+    rclpy.init(args)
+
+    node = rclpy.create_node('talker')
+
+    chatter_pub = node.create_publisher(String, 'chatter', qos_profile_default)
+
+    msg = String()
+
+    i = 1
+    while True:
+        msg.data = 'Hello World: {0}'.format(i)
+        i += 1
+        print('Publishing: "{0}"'.format(msg.data))
+        chatter_pub.publish(msg)
+        sleep(1)
+
+if __name__ == '__main__':
+    main()
--- a/rclpy_examples/listener_py.py
+++ b/rclpy_examples/listener_py.py
@@ -1,0 +1,40 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+import rclpy
+from rclpy.qos import qos_profile_default
+from std_msgs.msg import String
+
+
+def chatter_callback(msg):
+    print('I heard: [%s]' % msg.data)
+
+
+def main(args=None):
+    if args is None:
+        args = sys.argv
+
+    rclpy.init(args)
+
+    node = rclpy.create_node('listener')
+
+    sub = node.create_subscription(String, 'chatter', chatter_callback, qos_profile_default)
+    assert sub  # prevent unused warning
+
+    rclpy.spin(node)
+
+if __name__ == '__main__':
+    main()
--- a/rclpy_examples/package.xml
+++ b/rclpy_examples/package.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>rclpy_examples</name>
+  <version>0.0.0</version>
+  <description>Package containing examples of how to use the rclpy API.</description>
+  <maintainer email="esteve@osrfoundation.org">Esteve Fernandez</maintainer>
+  <license>Apache License 2.0</license>
+
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+
+  <test_depend>rclpy</test_depend>
+  <test_depend>std_msgs</test_depend>
+
+  <export>
+    <build_type>ament_python</build_type>
+  </export>
+</package>
--- a/rclpy_examples/setup.py
+++ b/rclpy_examples/setup.py
@@ -1,0 +1,29 @@
+from setuptools import setup
+
+setup(
+    name='rclpy_examples',
+    version='0.0.0',
+    packages=[],
+    py_modules=['listener_py', 'talker_py'],
+    install_requires=['setuptools'],
+    author='Esteve Fernandez',
+    author_email='esteve@osrfoundation.org',
+    maintainer='Esteve Fernandez',
+    maintainer_email='esteve@osrfoundation.org',
+    keywords=['ROS'],
+    classifiers=[
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python',
+        'Topic :: Software Development',
+    ],
+    description='Package containing examples of how to use the rclpy API.',
+    license='Apache License, Version 2.0',
+    test_suite='test',
+    entry_points={
+        'console_scripts': [
+            'listener_py = listener_py:main',
+            'talker_py = talker_py:main',
+        ],
+    },
+)
--- a/rclpy_examples/test/test_copyright.py
+++ b/rclpy_examples/test/test_copyright.py
@@ -1,0 +1,20 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ament_copyright.main import main
+
+
+def test_copyright():
+    rc = main(argv=['.', 'test'])
+    assert rc == 0, 'Found errors'
--- a/rclpy_examples/test/test_pep257.py
+++ b/rclpy_examples/test/test_pep257.py
@@ -1,0 +1,20 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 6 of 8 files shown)


rclpy_examples/talker_asyncio.py
```diff
@@ -0,0 +1,33 @@
...
+    msg = String()
+
+    i = 1
+    while rclpy.ok():
+      msg.data = "Hello World: " + i
+      i += 1
+      print("Publishing: '%s'" % msg.data)
+      chatter_pub.publish(msg)
+      yield from rclpy.spin_some(node)
+      yield from loop_rate.sleep()
```
wjwwood: Seems like we could do something a little more `asyncio`y here. Maybe like this:

```python
import asyncio
import sys

import rclpy
from std_msgs.msg import String


@asyncio.coroutine
def publish_loop(publisher, loop_rate):
    msg = String()
    i = 1
    while rclpy.ok():
        msg.data = "Hello World: {0}".format(i)
        i += 1
        print("Publishing: '{0}'".format(msg.data))
        publisher.publish(msg)
        yield from loop_rate.sleep()


@asyncio.coroutine
def main(args):
    rclpy.init(args)
    node = rclpy.create_node("talker")
    chatter_pub = node.create_publisher(String, "chatter", rclpy.qos_profile_default)
    loop_rate = rclpy.WallRate(2)
    asyncio.ensure_future(publish_loop(chatter_pub, loop_rate))
    yield from rclpy.spin(node)

if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main(sys.argv))
    loop.close()

```

The benefit here is that if `yield from rclpy.spin_some(node)` internally has `yield from`'s the `loop_rate.sleep()` can return between them. Basically we can avoid having to interleave the activities ourself and instead let asyncio do it, and more efficiently at that. You could also consider using `asyncio.wait` (https://docs.python.org/3/library/asyncio-task.html#asyncio.wait) in combination with `rclpy.spin(node)`, `loop_rate.sleep`, and `asyncio.FIRST_COMPLETED`.

rclpy_examples/talker_asyncio.py
```diff
@@ -0,0 +1,33 @@
...
+
+if __name__ == '__main__':
+    loop - asyncio.get_event_loop()
```
wjwwood: `=` not `-` (you should get a Python syntax checker for your editor if you don't already have one)

rclpy_examples/talker_asyncio.py
```diff
@@ -0,0 +1,33 @@
+import asyncio
+import sys
+
+import rclpy
+from rclpy.qos import qos_profile_default
```
wjwwood: This could probably live within `rclpy`.

rclpy_examples/listener_asyncio.py
```diff
@@ -0,0 +1,25 @@
...
+
+import rclpy
+from rclpy.qos import qos_profile_default
+from std_msgs.msg import String
+
+
+def chatter_callback(msg):
+    print("I heard: [%s]" % msg.data)
+
+@asincio.coroutine
```
wjwwood: nit pik: two lines between free functions, PEP8

rclpy_examples/talker_py.py
```diff
@@ -0,0 +1,29 @@
...
+
+    chatter_pub = node.create_publisher(String, "chatter", qos_profile_default)
+
+    loop_rate = rclpy.WallRate(2)
+
+    msg = String()
+
+    i = 1
+    while rclpy.ok():
+      msg.data = "Hello World: " + i
```
wjwwood: This will not work I think, should give an error like `No '+' operator between types 'str' and 'int'`. You'll need to do a formatted string `msg.data = "Hello World: {0}".format(i)`.

Btw, in Python3.6 we can use string interpolation! Like this: `msg.data = f"Hello World: {i}"` I'm so excited: https://www.python.org/dev/peps/pep-0498/

jacquelinekay:
Can you incorporate a `spin`/`spin_some`/`spin_once` call with a timeout somewhere, so that we know what that will look like?

rclpy_examples/talker_asyncio.py
esteve: Looks good, but I think I'm removing the `asyncio` examples from this PR, they are a bit distracting since the focus of this PR is more about the `rclpy` API. But thanks for the review :-)

wjwwood: Another way to write it:

```Python
import asyncio
import sys

import rclpy
from std_msgs.msg import String


@asyncio.coroutine
def publish_loop(publisher, loop_rate):
    msg = String()
    i = 1
    while rclpy.ok():
        msg.data = "Hello World: {0}".format(i)
        i += 1
        print("Publishing: '{0}'".format(msg.data))
        publisher.publish(msg)
        yield from loop_rate.sleep()


def main(args):
    rclpy.init(args)
    node = rclpy.create_node("talker")
    chatter_pub = node.create_publisher(String, "chatter", rclpy.qos_profile_default)
    loop_rate = rclpy.WallRate(2)
    asyncio.ensure_future(publish_loop(chatter_pub, loop_rate))
    loop = asyncio.get_event_loop()
    asyncio.ensure_future(rclpy.spin(node, on_shutdown=lambda: loop.close()))
    loop.run_forever()

if __name__ == '__main__':
    main(sys.argv)

```

Which could be simplified further with some syntactic sugar from `rclpy`:

```Python
import asyncio
import sys

import rclpy
from std_msgs.msg import String


@asyncio.coroutine
def publish_loop(publisher, loop_rate):
    msg = String()
    i = 1
    while rclpy.ok():
        msg.data = "Hello World: {0}".format(i)
        i += 1
        print("Publishing: '{0}'".format(msg.data))
        publisher.publish(msg)
        yield from loop_rate.sleep()


def main(args):
    rclpy.init(args)
    node = rclpy.create_node("talker")
    chatter_pub = node.create_publisher(String, "chatter", rclpy.qos_profile_default)
    loop_rate = rclpy.WallRate(2)
    asyncio.ensure_future(publish_loop(chatter_pub, loop_rate))
    rclpy.asyncio_spin(node, loop=asyncio.get_event_loop())

if __name__ == '__main__':
    main(sys.argv)

```

rclpy_examples/talker_py.py
esteve: Fixed using `.format()`. Literal string interpolation looks neat, indeed.

esteve:
@jacquelinekay besides `spin_until_future_complete`, I think we're not using `spin` / `spin_some` / spin_once` with timeouts anywhere else. I can't find an example where we use the `spin` functions with a timeout, unless services are involved, which use `spin_until_future_complete`, but we we don't support services in Python yet. Do you know if there's a place in our code where we use `spin` with timeouts, without `spin_until_future_complete`?

jacquelinekay:
It's a part of the rclcpp API for spin_once:
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/executor.hpp#L126

The default is a blocking spin.

A timeout value of 0 is non-blocking (see https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_multithreaded.cpp#L96 and https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_intra_process.cpp#L65).

`spin_once` with a timeout is tested here: https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_timeout_subscriber.cpp#L67

This is an important feature for real-time; there needs to be a guarantee that the execution thread does not block indefinitely.

esteve:
Thanks. I'd argue that supporting real time in Python would be difficult, since there are many memory allocations in the Python interpreter that can't be controlled. However, once we introduce services and use futures, we might need to add support for timeouts in `spin` and connect it to Python future's `result(timeout=timeout_value)`, though I don't think we need timeouts now for plain `publish` / `subscribe`

gbiggs:
Real-time coding in Python probably isn't a reality now, but there are efforts to create real-time capable Python implementations (such as Micro Python). I haven't used any of them so I can't speak for how good they are or if they'll ever succeed, but I think that unless the effort is too great or it would make the implementation un-Pythonic, it is worth at least giving some consideration to designing a Python API that doesn't _prevent_ real-time, assuming the underlying interpreter/libraries/whatever provides support.

rclpy_examples/talker_py.py
```diff
@@ -0,0 +1,32 @@
...
+
+    msg = String()
+
+    i = 1
+    while rclpy.ok():
+      msg.data = "Hello World: {0}".format(i)
+      i += 1
+      print("Publishing: '%s'" % msg.data)
+      chatter_pub.publish(msg)
+      rclpy.spin_some(node)
```
dirk-thomas: Will this `spin_some` call be necessary?

jacquelinekay:
There are some commits mixed into this branch that affect `rclc_examples`. I think @wjwwood suggested we not incorporate those examples into master for now. Should we remove those commits from this branch?

esteve:
@jacquelinekay yeah, makes sense, I'll remove them.

esteve:
@jacquelinekay removed.

dirk-thomas:
Nitpik: single quotes.

esteve:
@dirk-thomas done.

Metanitpicking: i's nitpick, not nitpik :stuck_out_tongue_winking_eye:

gerkey:
When I switch to these branches to try things out on my OSX machine, I get an error when trying to do a non-isolated symlink build:

~~~
Process package 'rclpy_examples' with context:
--------------------------------------------------------------------------------
 source_space => /Users/gerkey/ros2_ws/src/ros2/examples/rclpy_examples
  build_space => /Users/gerkey/ros2_ws/build/rclpy_examples
install_space => /Users/gerkey/ros2_ws/install
   make_flags => -j4, -l4
  build_tests => True
--------------------------------------------------------------------------------
+++ Building 'rclpy_examples'
+++ Installing 'rclpy_examples'
-- [ament] Deploying: /Users/gerkey/ros2_ws/install/share/rclpy_examples/package.xml
-- [ament] Deploying: /Users/gerkey/ros2_ws/install/share/rclpy_examples/environment/path.sh
-- [ament] Deploying: /Users/gerkey/ros2_ws/install/share/rclpy_examples/environment/pythonpath.sh
-- [ament] Deploying: /Users/gerkey/ros2_ws/install/share/rclpy_examples/local_setup.bash
-- [ament] Deploying: /Users/gerkey/ros2_ws/install/share/rclpy_examples/local_setup.sh
-- [ament] Deploying: /Users/gerkey/ros2_ws/install/share/rclpy_examples/local_setup.zsh
Traceback (most recent call last):
  File "./src/ament/ament_tools/scripts/ament.py", line 149, in <module>
    sys.exit(main() or 0)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/commands/ament.py", line 88, in main
    rc = args.main(args)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/verbs/build/cli.py", line 152, in main
    iterate_packages(opts, packages, per_package_main)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/verbs/build/cli.py", line 255, in iterate_packages
    rc = per_package_callback(opts)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/verbs/build_pkg/cli.py", line 271, in main
    return run(opts, context)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/verbs/build_pkg/cli.py", line 323, in run
    handle_build_action(on_install_ret, context)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/verbs/build_pkg/cli.py", line 263, in handle_build_action
    build_action.cmd(context)
  File "/Users/gerkey/ros2_ws/src/ament/ament_tools/ament_tools/build_types/ament_python.py", line 338, in _install_action_python
    os.symlink(src, dst)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/gerkey/ros2_ws/src/ros2/examples/rclpy_examples/' -> '/Users/gerkey/ros2_ws/build/rclpy_examples/'
~~~

Indeed, `/Users/gerkey/ros2_ws/build/rclpy_examples/` does not exist. I tried creating it manually then running the build again, but the error is repeated and along the way the manually created directory is deleted (presumably as part of the cleanup that happens before starting to build the package).

The problem goes away if I do a non-symlink build (remove the `-s` arg from the call to `ament.py`).

Is that an issue with the configuration of this package, or does it point to something wrong in ament?

dirk-thomas:
The problems looks to be related to an incorrect setup.py file. Here https://github.com/ros2/examples/pull/87/files#diff-311de9e783d1c635671d2e11e1c71122R7 the setup.py files declares one package but the folder doesn't contain a Python package anywhere?

If I install the Python package manually at actually warns explicitly that '' is not a valid package:

    WARNING: '' not a valid package name; please use only.-separated package names in setup.py

I assume you got the some message when you build without `--symlinked-install`?

Beside fixing the `setup.py` file we can somehow make ament handle this more gracefully. I am not sure if a warning is the right thing to do (to mimic setuptools) or an error would be more helpful for the user?

gerkey:
For those of us not so familiar with setup.py files, what's the required fix?

dirk-thomas:
Since the ament package does not contain a Python package I would guess an empty list should be passed. Also all other ROS 2 packages use `find_packages(...)` to automatically determine the Python packages. I assume that would also work and would be the preferred solution.

dirk-thomas:
The setup file stills needs to be updated. Moving back to in-progress.

esteve:
@dirk-thomas I've updated the setup.py file.

dirk-thomas:
Is the warning gone now. Then @gerkey use case should work fine too.

Should we add the Python linter to this package?

PR #88, opened 2016-02-05
C typesupport tests: connext

jacquelinekay:
See ros2/rcl#19


wjwwood:
I made a style fix up, but I think we can roll this in.

PR #92, opened 2016-02-26
Use target suffix for rclcpp

### LINKED ISSUE #91: Fix linking of examples
Use the pattern in:

https://github.com/ros2/demos/blob/master/intra_process_demo/CMakeLists.txt#L33

instead of:

https://github.com/ros2/examples/blob/aeca79cdfc0d0263adff922f01df65ab367b87d5/rclcpp_examples/CMakeLists.txt#L21

esteve:
Fixes #91


dirk-thomas:
Absolutely. +1

PR #93, opened 2016-03-04
wait for service in client example

gerkey:
The test that uses this client example can fail if the server isn't yet up and ready to hear the client's request. In that situation, the client waits forever and the test times out, as we've [seen](http://ci.ros2.org/view/nightly/job/nightly_osx/234/testReport/junit/projectroot.Users.osrf.jenkins.workspace.nightly_osx.ws.src.ros2.examples/rclcpp_examples/test_example_add_two_ints_server_add_two_ints_client__rmw_fastrtps_cpp/) with FastRTPS. This PR adds a `wait_for_service()` call that the client uses to delay its request until the time is right.

This function should not live here in the long term, but rather be improved and promoted into client API (ros2/rclcpp#201). But this is a decent place to park it for now.

### DIFF
--- a/rclcpp_examples/src/services/add_two_ints_client.cpp
+++ b/rclcpp_examples/src/services/add_two_ints_client.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <iostream>
+#include <string>
 
 #include "rclcpp/rclcpp.hpp"
 
@@ -35,6 +36,47 @@ example_interfaces::srv::AddTwoInts_Response::SharedPtr send_request(
   }
 }
 
+// This function could be considered a prototype of an API that should be
+// promoted into rclcpp. This version polls; doing it in an event-driven
+// manner would be better.
+bool wait_for_service(rclcpp::Node::SharedPtr node,
+  const std::string & topic,
+  const std::chrono::steady_clock::duration timeout)
+{
+  const std::chrono::milliseconds sleep_per_loop(1);
+  auto end = std::chrono::steady_clock::now() + timeout;
+  bool success = false;
+  bool printed = false;
+  while (std::chrono::steady_clock::now() < end) {
+    try {
+      if (node->count_subscribers(topic) == 0) {
+        success = true;
+        break;
+      }
+    } catch (std::runtime_error & e) {
+      // fastrtps doesn't yet implement rmw_count_subscribers(), but rather
+      // throws an exception instead. We'll just run out the clock and
+      // return true, hoping for the best.
+      if (!printed) {
+        fprintf(stderr, "count_subscribers() exception (this is expected for FastRTPS): %s\n",
+          e.what());
+        printed = true;
+      }
+      rmw_reset_error();
+      success = true;
+    }
+    auto remaining = end - std::chrono::steady_clock::now();
+    if (remaining < sleep_per_loop) {
+      if (remaining > std::chrono::steady_clock::duration::zero()) {
+        std::this_thread::sleep_for(remaining);
+      }
+    } else {
+      std::this_thread::sleep_for(sleep_per_loop);
+    }
+  }
+  return success;
+}
+
 int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
@@ -45,6 +87,11 @@ int main(int argc, char ** argv)
   auto request = std::make_shared<example_interfaces::srv::AddTwoInts::Request>();
   request->a = 2;
   request->b = 3;
+
+  if (!wait_for_service(node, "add_two_ints", std::chrono::seconds(2))) {
+    printf("add_two_ints service not available. Exiting.\n");
+    return 1;
+  }
 
   // TODO(wjwwood): make it like `client->send_request(node, request)->sum`
   // TODO(wjwwood): consider error condition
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Since the problem probably affects all of the service tests does it make sense to only patch it for this single one?

gerkey:
* http://ci.ros2.org/job/ci_linux/1059/
* http://ci.ros2.org/job/ci_osx/869/
* Windows is in the queue

gerkey:
@dirk-thomas The right fix is to implement the `wait_for_service` call in the client library and use it everywhere, including here. My implementation isn't yet ready to be promoted in that way (we may need some changes to the rmw API), and I don't want to copy this code everywhere that services are called in our tests. If, before we address ros2/rclcpp#201, we see other failures caused by the same underlying problem, I'll look at factoring it out for reuse, at least within this package.

rclcpp_examples/src/services/add_two_ints_client.cpp
```diff
@@ -35,6 +36,41 @@ example_interfaces::srv::AddTwoInts_Response::SharedPtr send_request(
...
+        success = true;
+        break;
+      }
+    } catch (std::runtime_error & e) {
+      // fastrtps doesn't yet implement rmw_count_subscribers(), but rather
+      // throws an exception instead. We'll just run out the clock and
+      // return true, hoping for the best.
+      rmw_reset_error();
+      success = true;
+    }
```
wjwwood: This is a fairly blanket error case, maybe you could either check the error message contents or at least print it out so that when there is an error that doesn't match what you expected you can see it in the log? That might break your example test though.

wjwwood:
+1 The change makes sense to me. It's unfortunate that FastRTPS doesn't implement the solution I suggested to you :sad:.

gerkey:
This change made more tests fail on Linux and it didn't even build on Windows. Fun!

gerkey:
Windows build fixed and same logic added to async client (it'll be removed from both examples once we resolve ros2/rclcpp#201). New CI:

* http://ci.ros2.org/job/ci_windows/1167/
* http://ci.ros2.org/job/ci_linux/1100/
* http://ci.ros2.org/job/ci_osx/885/

gerkey:
CI shows only failures and warning that are unrelated to this change.

gerkey:
@wjwwood You +1ed previously, but can you have a second look before I merge?

rclcpp_examples/src/services/add_two_ints_client.cpp
```diff
@@ -35,6 +36,47 @@ example_interfaces::srv::AddTwoInts_Response::SharedPtr send_request(
...
+  auto end = std::chrono::steady_clock::now() + timeout;
+  bool success = false;
+  bool printed = false;
+  while (std::chrono::steady_clock::now() < end) {
+    try {
+      if (node->count_subscribers(topic) == 0) {
+        success = true;
+        break;
+      }
+    } catch (std::runtime_error & e) {
```
wjwwood: One issue here is that `node->count_subscribers` should never throw a C++ exception. This is actually a problem with the fast rtps implementation, but ideally it would catch the exception internally and convert it to a return code.

rclcpp_examples/src/services/add_two_ints_client.cpp
```diff
@@ -35,6 +36,47 @@ example_interfaces::srv::AddTwoInts_Response::SharedPtr send_request(
...
+  const std::string & topic,
+  const std::chrono::steady_clock::duration timeout)
+{
+  const std::chrono::milliseconds sleep_per_loop(1);
+  auto end = std::chrono::steady_clock::now() + timeout;
+  bool success = false;
+  bool printed = false;
+  while (std::chrono::steady_clock::now() < end) {
+    try {
+      if (node->count_subscribers(topic) == 0) {
```
wjwwood: Maybe I'm reading the logic wrong, but shouldn't you be checking for count_subscribers to be greater than zero?

rclcpp_examples/src/services/add_two_ints_client.cpp
wjwwood: I should have said, it should never throw a C++ exception because it is a C function.

gerkey: The underlying function `rmw_count_subscribers()` is C (for fastrtps, it's hardcoded to return an error: https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/blob/master/rmw_fastrtps_cpp/src/functions.cpp#L1793-L1801), but here I'm calling a C++ class method: https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node.cpp#L317-L329.

rclcpp_examples/src/services/add_two_ints_client.cpp
gerkey: Uh....yeah. That's totally backward and broken. Please hold.

rclcpp_examples/src/services/add_two_ints_client.cpp
wjwwood: Ah, my bad. Sorry for the noise :bell:.

gerkey:
When the logic is fixed to actually require that `> 0` subscribers are hooked up, the test fails for opensplice and both flavors of connext. I suspect that the problem is that, as @wjwwood previously suggested, you can't assume that services are implemented atop topics and check for subscribers to infer that a service server is up. This approach is a dead end.

PR #94, opened 2016-03-09
temporary try/except block around import rclpy

jacquelinekay:
Connects to ros2/rosidl#99

This is a temporary workaround for the failing nosetest for these examples until rclpy works with all middlewares (ros2/rclpy#5).

### DIFF
--- a/rclpy_examples/listener_py.py
+++ b/rclpy_examples/listener_py.py
@@ -14,7 +14,13 @@
 
 import sys
 
-import rclpy
+# TODO(jacquelinekay): Remove try block when rclpy supports multiple vendors!!
+try:
+    import rclpy
+except ImportError:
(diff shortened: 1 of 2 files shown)


rclpy_examples/listener_py.py
```diff
@@ -14,7 +14,13 @@
 
 import sys
 
-import rclpy
+# TODO(jacquelinekay): Remove try block when rclpy supports multiple vendors!!
+try:
+    import rclpy
+except ImportError:
+    print("rclpy was not found. This could be because no valid typesupport was found.")
+    exit()
```
wjwwood: It doesn't really matter, especially since it is temporary, but `sys.exit` is probably better to use here: http://stackoverflow.com/a/6501134/671658

PR #104, opened 2016-04-06
Remove unused rmw typesupport call

jacquelinekay:
While working on ros2/rmw#63 I noticed that this call is never used, so I'm removing it (similar to ros2/demos#63).


dirk-thomas:
+1

PR #105, opened 2016-04-14
Test rmw linking of rclcpp examples

dhood:
Regression test for https://github.com/ros2/examples/commit/fda593878912e0173861c44d9600396b3e6a087f. If https://github.com/ros2/examples/commit/fda593878912e0173861c44d9600396b3e6a087f is reverted, the `test_output` tests pass, but the `test_linking` tests fail (on windows 10, trusty, and yosemite). 

todo: 
- [x] check linking of default executables against default rmw implementation

### DIFF
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -18,7 +18,7 @@ function(custom_executable subfolder target)
   add_executable(${target}${target_suffix}
     src/${subfolder}/${target}.cpp)
   ament_target_dependencies(${target}${target_suffix}
-    "rclcpp"
+    "rclcpp${target_suffix}"
     "std_msgs"
     "example_interfaces")
   install(TARGETS ${target}${target_suffix}
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -53,18 +53,6 @@ macro(targets)
   custom_executable(parameters set_and_get_parameters_async)
 endmacro()
 
-set(executable_list
-    list_parameters_async
-    list_parameters
-    parameter_events_async
-    parameter_events
-    set_and_get_parameters_async
-    set_and_get_parameters
-    talker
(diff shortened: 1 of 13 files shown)


rclcpp_examples/CMakeLists.txt
```diff
@@ -59,52 +59,88 @@ if(AMENT_ENABLE_TESTING)
...
+      parameter_events_async
+      parameter_events
+      set_and_get_parameters_async
+      set_and_get_parameters
+      talker
+      listener
+      add_two_ints_server
+      add_two_ints_client
+      add_two_ints_client_async)
+
```
dhood: is there anything I can call to access a list of all executables instead of writing them out myself?
I wanted to append names to a list when `custom_executable` is called, but I'm having issues with scope

dirk-thomas: You could reuse the list already in the CMake code as `example_output_tests`.

wjwwood: Even though `custom_executable` is a `function`, you should be able to do `list(APPEND ...)` on a variable already set outside of the function. The other option is that you can do `set(x ${x} PARENT_SCOPE)`, where `x` is the variable you want to export up one level (out of a function for example).

dirk-thomas: `custom_executable` is a function and therefore doesn't (normally) affect the callers scope. What are you trying to achieve.

dhood: I was trying to use a list separate to `example_output_tests` since there might be things you want to check the linking of but not the output.

I wasn't able to do the `list(APPEND` inside `custom_executable` even with `PARENT_SCOPE` because it's called from within `call_for_each_rmw_implementation`, so things were starting to get messy.

tfoote:
We think that https://github.com/ros2/examples/pull/106 is sufficient. We'll close this and reopen it from your fork if necessary.

Issue #96, opened 2016-03-22
Verify that the rmw implementation matches executable


jacquelinekay:
@dhood will work on making sure the name of the executable actually matches the middleware implementation.

wjwwood:
@dhood I just re-reviewed all the associated PR's. I'd say, run one more set of CI on it (just to make sure something weird didn't happen) and if they pass merge them. Thanks for working on it!

dhood:
yep it's already running now thanks!

dhood:
http://ci.ros2.org/job/ci_osx/1044/ (no failures)
http://ci.ros2.org/job/ci_windows/1348/ (failures with age >= 1)
http://ci.ros2.org/job/ci_linux/1323/ (no failures)

merging these prs now

PR #106, opened 2016-04-26
Verify rmw implementation of rclcpp examples at runtime

dhood:
Modifies existing `rclcpp_examples` tests to act as a regression test for https://github.com/ros2/examples/commit/fda593878912e0173861c44d9600396b3e6a087f (in comparison to #105 which adds new tests to explicitly check the linking)

Currently just for `talker` - once approved I can extend to the other executables

Note that I had to modify the exit handler used in the tests in order to get them to fail with the non-zero return code from `talker` (when https://github.com/ros2/examples/commit/fda593878912e0173861c44d9600396b3e6a087f is reverted). This does not seem to have any negative side-effects (no regressions in CI). As a result, I am not entirely clear on the reason it was `ignore_exit_handler` before, given that it suppresses errors and seems to have been unnecessary. So, please be extra skeptical of [this change](https://github.com/dhood/examples/commit/27aa2164156d06d54ae8c2cd5596d1215544db6d#diff-1988d847d73fdc4f80e5573bb07a0466R31) in this PR.

http://ci.ros2.org/job/ci_windows/1265/
http://ci.ros2.org/job/ci_linux/1218/
http://ci.ros2.org/job/ci_osx/983/

Connects to ros2/examples#96

### DIFF
--- a/rclcpp_examples/package.xml
+++ b/rclcpp_examples/package.xml
@@ -24,7 +24,6 @@
   <exec_depend>sensor_msgs</exec_depend>
   <exec_depend>std_msgs</exec_depend>
 
-  <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_cmake_nose</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -98,7 +98,8 @@ if(AMENT_ENABLE_TESTING)
 
       ament_add_nose_test(test_example_${exe_list_underscore}${target_suffix}
         "${CMAKE_CURRENT_BINARY_DIR}/test_${exe_list_underscore}${target_suffix}_$<CONFIGURATION>.py"
-        TIMEOUT 30)
+        TIMEOUT 30
+        ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
       set_tests_properties(test_example_${exe_list_underscore}${target_suffix}
         PROPERTIES DEPENDS "test_example_${exe_list_underscore}${target_suffix} test_example_${exe_list_underscore}${target_suffix}")
     endforeach()


rclcpp_examples/package.xml
```diff
@@ -24,7 +24,6 @@
   <exec_depend>sensor_msgs</exec_depend>
   <exec_depend>std_msgs</exec_depend>
 
-  <test_depend>ament_cmake_gtest</test_depend>
```
wjwwood: I take it that gtest is not used anywhere?

dhood: not that I can see. I ran CI to be sure

wjwwood: 👍

rclcpp_examples/src/topics/talker.cpp
```diff
@@ -13,13 +13,28 @@
...
+  #ifdef EXPECTED_RMW_IMPLEMENTATION_IDENTIFIER
+  std::string rmw_identifier(rmw_get_implementation_identifier());
+
+  if (rmw_identifier.compare(STRING(EXPECTED_RMW_IMPLEMENTATION_IDENTIFIER)) != 0) {
+    std::cerr << "Expected RMW implementation identifier of '" <<
+      STRING(EXPECTED_RMW_IMPLEMENTATION_IDENTIFIER) << "' but instead found '" <<
+      rmw_identifier << "'" << std::endl;
+    return 1;
+  }
+  #endif
```
wjwwood: This is the kind of thing that I didn't want to have in the examples. (I think it detracts from the otherwise simple example code)

I was thinking we could instead put the ability to check this into the `rmw` library itself. I never prototyped this, but I'm still interested in doing it.

If we merge this as-is, I'd want a TODO to remind us that there is probably a way to do it which doesn't affect the user code here.

wjwwood: @dhood see: https://github.com/ros2/rcl/pull/51

dirk-thomas:
Before our executable did not exit cleanly when they received a `SIGINT`. Therefore the tests needed to ignore their exit code using the `ignore_exit_handler`. Maybe that has changed in the mean time.

rclcpp_examples/src/topics/talker.cpp
dhood: thanks - I'll integrate it

tfoote:
@dhood "should we have a macro that checks this for all tests which call it?": no, just opt in where we think it is needed.

Issue #107, opened 2016-04-26
compiler warning operator delete


dirk-thomas:
Using g++ 5.3 on Xenial the compiler warns about the following:

```
.../src/ros2/examples/rclcpp_examples/src/topics/allocator_example.cpp:141:6: warning: ‘void operator delete(void*, size_t)’ is a usual (non-placement) deallocation function in C++14 (or with -fsized-deallocation) [-Wc++14-compat]
 void operator delete(void * ptr, size_t) noexcept
      ^
.../src/ros2/examples/rclcpp_examples/src/topics/allocator_example.cpp:141:6: warning: ‘void operator delete(void*, size_t)’ is a usual (non-placement) deallocation function in C++14 (or with -fsized-deallocation) [-Wc++14-compat]
 void operator delete(void * ptr, size_t) noexcept
      ^
.../src/ros2/examples/rclcpp_examples/src/topics/allocator_example.cpp:141:6: warning: ‘void operator delete(void*, size_t)’ is a usual (non-placement) deallocation function in C++14 (or with -fsized-deallocation) [-Wc++14-compat]
 void operator delete(void * ptr, size_t) noexcept
      ^
.../src/ros2/examples/rclcpp_examples/src/topics/allocator_example.cpp:141:6: warning: ‘void operator delete(void*, size_t)’ is a usual (non-placement) deallocation function in C++14 (or with -fsized-deallocation) [-Wc++14-compat]
 void operator delete(void * ptr, size_t) noexcept
      ^
```

PR #109, opened 2016-06-07
Clean up allocator for gcc 5.3

jacquelinekay:
Connects to #107 

**Do not merge until we update to Xenial; this will break the build with the default gcc on Trusty.**


jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_linux/1411/)

dirk-thomas:
Should the branch name be changed to `xenial` so that the patch can be tested together with all the other changes?

dirk-thomas:
I also deleted the xenial branch.

jacquelinekay:
Thanks.

PR #110, opened 2016-06-09
use wait_for_service to make examples less flaky

wjwwood:
Connects to ros2/ros2#215

### DIFF
--- a/rclcpp_examples/CMakeLists.txt
+++ b/rclcpp_examples/CMakeLists.txt
@@ -85,10 +85,8 @@ if(BUILD_TESTING)
     foreach(example_test ${example_tests_to_test})
       string(REPLACE ":" ";" example_executables "${example_test}")
       set(RCLCPP_EXAMPLES_EXPECTED_OUTPUT "")
-      set(example_executable_targets "")
       foreach(executable ${example_executables})
         list(APPEND RCLCPP_EXAMPLES_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/${executable}")
-        list(APPEND example_executable_targets ${executable}${target_suffix})
       endforeach()
 
       set(RCLCPP_EXAMPLES_EXECUTABLE "")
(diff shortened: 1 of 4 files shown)


rclcpp_examples/CMakeLists.txt
```diff
@@ -68,12 +68,21 @@ if(BUILD_TESTING)
...
+    "talker:listener")
+  set(fastrtps_excluded_example_tests
     "add_two_ints_server:add_two_ints_client"
     "add_two_ints_server:add_two_ints_client_async")
 
   macro(tests)
-    foreach(example_test ${example_tests})
+    set(example_tests_to_test ${example_tests})
+    # TODO(wjwwood): recombine with example_tests this when fastrtps supports wait_for_service.
+    if("${rmw_implementation} " STREQUAL "rmw_fastrtps_cpp ")
```
dirk-thomas: Since we are using CMake 3.5 now please avoid the trailing spaces as well as the quotes and explicit variable expansion if the variable is know to exist:

    if(rmw_implementation STREQUAL "rmw_fastrtps_cpp")

wjwwood: 61590ca

rclcpp_examples/CMakeLists.txt
```diff
@@ -68,12 +68,21 @@ if(BUILD_TESTING)
...
     "add_two_ints_server:add_two_ints_client_async")
 
   macro(tests)
-    foreach(example_test ${example_tests})
+    set(example_tests_to_test ${example_tests})
+    # TODO(wjwwood): recombine with example_tests this when fastrtps supports wait_for_service.
+    if(rmw_implementation STREQUAL "rmw_fastrtps_cpp")
+      message(STATUS "Skipping some example tests for '${rmw_implementation}'")
+    else()
+      set(example_tests_to_test ${example_tests} ${fastrtps_excluded_example_tests})
```
dirk-thomas: This should `list(APPEND` instead of repeating the content from line 77.

dirk-thomas:
Beside the comment +1

rclcpp_examples/CMakeLists.txt
wjwwood: I guess it could, is this the right syntax? 4e2d944

dirk-thomas: Yes.

PR #116, opened 2016-07-09
Add ros2param

gerkey:
This is an early prototype of a replacement for `rosparam`. As we've discussed, we'll eventually want this written in Python, but until then, it's incredibly handy to have a working tool to experiment with. I had to pick a command-line syntax, so I went with `node/param` to indicate that we're talking about `param` owned by `node`. That's certainly up for debate, but I'm not aiming to settle the syntax once and for all here. Rather, I see this experiment as a way to try out different ideas before having to commit a particular way of doing things.

I was planning to write a test in this package for this tool, but I ran out of time. If the consensus is that the tools shouldn't be added here without a test, I'll drop it over in https://github.com/ros2/turtlebot2_demo for now.

Btw, over there, we now have an example of a real node reacting to dynamic parameter changes: https://github.com/ros2/turtlebot2_demo/blob/master/joy/joy_node.cpp#L36-L62. We can change the linear and angular scales for joystick control on the fly, without extra machinery! The usage is still a little clunky; I'll be submitting some PRs to improve the parameter APIs (e.g., https://github.com/ros2/rclcpp/issues/238).

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1650/) 
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1239/)

### DIFF
--- a/rclcpp_examples/src/parameters/ros2param.cpp
+++ b/rclcpp_examples/src/parameters/ros2param.cpp
@@ -1,0 +1,132 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+#include "rclcpp/rclcpp.hpp"
+
+#define USAGE "USAGE:\n  ros2param get <node/variable>\n  ros2param set <node/variable> <value>"
+
+typedef enum
+{
+  PARAM_NONE,
+  PARAM_GET,
+  PARAM_SET,
+} param_operation_t;
+
+rclcpp::parameter::ParameterVariant
+parse_args(int argc, char ** argv, std::string & remote_node, param_operation_t & op)
+{
+  if (argc < 3) {
+    return rclcpp::parameter::ParameterVariant();
+  }
+
+  std::string verb = argv[1];
+  std::string name = argv[2];
+  size_t slash = name.find('/');
+  if ((slash == std::string::npos) ||
+    (slash == 0) ||
+    (slash == (name.size() - 1)))
+  {
+    return rclcpp::parameter::ParameterVariant();
+  }
+  remote_node = name.substr(0, slash);
+  std::string variable = name.substr(slash + 1, name.size() - slash - 1);
+
+  if ((verb == "get") && (argc == 3)) {
+    op = PARAM_GET;
+    return rclcpp::parameter::ParameterVariant(variable, 0);
+  } else if ((verb == "set") && (argc == 4)) {
+    op = PARAM_SET;
+    std::string value = argv[3];
+    char * endptr;
+    int l = strtol(value.c_str(), &endptr, 10);
+    if ((errno == 0) && (*endptr == '\0')) {
+      return rclcpp::parameter::ParameterVariant(variable, l);
+    } else {
+      errno = 0;
+      double d = strtod(value.c_str(), &endptr);
+      if ((errno == 0) && (*endptr == '\0')) {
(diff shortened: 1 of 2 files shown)


wjwwood:
Cool. I tried it out locally, but I wasn't able to set a parameter on just the `talker` node. Does the node have to already have "created" the parameter by setting it initially or can you set a parameter that the node doesn't read or use? Also, I was using Connext static because that's what I had handy so that might be the issue. I didn't try FastRTPS or OpenSplice.

I also, found myself wanting a "list" verb.

If it helps with your demo, as-is, then I think it lgtm.

gerkey:
Thanks for the feedback. I had only tested with OpenSplice, and it worked fine. As usual, there's going to be some difference with other implementations. I'm collecting extensions in #117 for (near) future work.

I'm going to merge so that this is in place for the alpha 7 release for use with the tbot demo.

rclcpp_examples/src/parameters/ros2param.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+  rclcpp::init(argc, argv);
+
+  std::string remote_node;
+  param_operation_t op = PARAM_NONE;
+
+  auto var = parse_args(argc, argv, remote_node, op);
+
+  if (op == PARAM_NONE) {
+    std::cout << USAGE << std::endl;
```
dirk-thomas: Instead of using `cout << ...` this should use `printf` (https://github.com/ros2/ros2/wiki/Developer-Guide#c-specific).

Same below.

rclcpp_examples/src/parameters/ros2param.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  if (op == PARAM_GET) {
+    auto get_parameters_result = parameters_client->get_parameters({var.get_name()});
+    auto get_result = rclcpp::spin_until_future_complete(
+      node, get_parameters_result, std::chrono::milliseconds(1000));
+    if ((get_result != rclcpp::executor::FutureReturnCode::SUCCESS) ||
+      (get_parameters_result.get().size() != 1) ||
+      (get_parameters_result.get()[0].get_type() == rclcpp::parameter::PARAMETER_NOT_SET))
+    {
+      std::cout << "Failed to get parameter" << std::endl;
```
dirk-thomas: Error messages should go to `stderr` (https://github.com/ros2/ros2/wiki/Developer-Guide#programming-conventions).

Same below.

rclcpp_examples/src/parameters/ros2param.cpp
```diff
@@ -0,0 +1,133 @@
...
+        } else if ((value == "false") || (value == "False")) {
+          return rclcpp::parameter::ParameterVariant(variable, false);
+        } else {
+          return rclcpp::parameter::ParameterVariant(variable, value);
+        }
+      }
+    }
+  } else {
+    return rclcpp::parameter::ParameterVariant();
+  }
```
dirk-thomas: To avoid the code to be nested (in this case four levels) I would suggest to utilize that the code returns in many locations. The following code doesn't need to be nested in an else block which reduces the overall nesting (and I think improves the readability since as a reader I don't have to consider all conditions of the parent blocks). For example like this (which has only two levels):

```
  if ((verb == "get") && (argc == 3)) {
    op = PARAM_GET;
    return rclcpp::parameter::ParameterVariant(variable, 0);
  }
  if ((verb == "set") && (argc == 4)) {
    op = PARAM_SET;
    std::string value = argv[3];
    char * endptr;
    int l = strtol(value.c_str(), &endptr, 10);
    if ((errno == 0) && (*endptr == '\0')) {
      return rclcpp::parameter::ParameterVariant(variable, l);
    }
    errno = 0;
    double d = strtod(value.c_str(), &endptr);
    if ((errno == 0) && (*endptr == '\0')) {
      return rclcpp::parameter::ParameterVariant(variable, d);
    }
    if ((value == "true") || (value == "True")) {
      return rclcpp::parameter::ParameterVariant(variable, true);
    }
    if ((value == "false") || (value == "False")) {
      return rclcpp::parameter::ParameterVariant(variable, false);
    }
    return rclcpp::parameter::ParameterVariant(variable, value);
  }
  return rclcpp::parameter::ParameterVariant();
```

gerkey:
@dirk-thomas Thanks for the feedback! Could you please open a new issue with those suggestions? I'm afraid that I'll lose track of comments on a closed PR.

dirk-thomas:
Please see #118.

Issue #125, opened 2016-10-04
install rclc examples


Karsten1987:
There is currently an `AMENT_IGNORE` file in the rclc examples due to a name clash with rclcpp i.e. 'talker' and 'listener'. 

Does it make sense to introduce a specific suffix for them? Such as `__rmw_fastrtps_c` ?

dirk-thomas:
I don't think the package is disabled because of that name clash.

But before it gets reenabled in the future the executables should be renamed to `talker_c` (plus the rmw suffix) similar to the Python equivalent.

The rmw suffix is defined by the rmw implementation package and there is no `rmw_fastrtps_c`.

codebot:
:hamburger: closed due to #126

PR #126, opened 2016-10-04
(fix) rename listener and talker with suffix _c

Karsten1987:
addresses https://github.com/ros2/examples/issues/125

### DIFF
--- a/rclc_examples/CMakeLists.txt
+++ b/rclc_examples/CMakeLists.txt
@@ -9,9 +9,10 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(example_interfaces REQUIRED)
 find_package(rclc REQUIRED)
-find_package(rclcpp REQUIRED)
+find_package(rcl REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
A CI build doesn't make sense since the package folder has an `AMENT_IGNORE` file. Did you build it locally? If yes, this looks good to me.

Karsten1987:
the rclc examples don't have much value right now since there is no real client library for c yet.

mikaelarguedas:
it doesn't compile locally because it's using outdated TYPESUPPORT macros and rely on code that doesn't exist (rclc is a dummy package with only header definition and no actual functions and no target created). So we may want to postpone this until there is an actual implementation of rclc

dirk-thomas:
While the package isn't "ready" yet I still think the proposed patch makes sense. I would be in favor of merging it.

Karsten1987:
we should then iterate over the complete cmake file and verify their dependencies.

Karsten1987:
I'll merge this due to its consistency with other client libraries. https://github.com/ros2/examples/blob/master/rclcpp_examples/CMakeLists.txt

We iterate over this once rclc is actually having example code.

Issue #124, opened 2016-10-04
[rclpy_examples] generate rmw_specific executables


mikaelarguedas:
Right now the only way to test different rmw_implementations is to populate the environment variable RCLPY_IMPLEMENTATION

wjwwood:
@mikaelarguedas can you move this one too? Thanks!

mikaelarguedas:
moved to https://github.com/ros2/tutorials/issues/4

PR #127, opened 2016-10-17
generate executable for each rmw_implementation

mikaelarguedas:
fixes https://github.com/ros2/examples/issues/124
connects to https://github.com/ros2/examples/issues/124

needs https://github.com/ament/ament_tools/issues/109 to be fixed before merge

### DIFF
--- a/rclpy_examples/listener_py.py
+++ b/rclpy_examples/listener_py.py
@@ -19,6 +19,7 @@ from rclpy.qos import qos_profile_default
 
 from std_msgs.msg import String
 
+import os
 
 def chatter_callback(msg):
     print('I heard: [%s]' % msg.data)
@@ -27,8 +28,12 @@ def chatter_callback(msg):
 def main(args=None):
     if args is None:
         args = sys.argv
+    else:
+        env = dict(os.environ)
+        env['RCLPY_IMPLEMENTATION'] = args
+        os.environ = env
 
-    rclpy.init(args)
+    rclpy.init()
 
     node = rclpy.create_node('listener')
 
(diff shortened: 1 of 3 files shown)


rclpy_examples/listener_py.py
```diff
@@ -27,16 +27,28 @@ def chatter_callback(msg):
...
+        String, 'chatter', chatter_callback, qos_profile_default)
     assert sub  # prevent unused warning
 
     while rclpy.ok():
         rclpy.spin_once(node)
 
+
+class main_for_rmw_impl_class(object):
+    def __getattr__(self, key):
+        return main([key])
```
dirk-thomas: `sys.argv` which the `args` keyword argument defaults to has a different semantic. It contains the script name as the first entry. Therefore I think this needs to be refactored somehow to have the same semantic when being invoked from the command line as well as the entry point.

mikaelarguedas:
good points, thanks.
fixed in https://github.com/ros2/examples/pull/127/commits/f59844f44aea4819e595e1a857592776426fb50e

rclpy_examples/listener_py.py
```diff
@@ -25,23 +25,20 @@ def chatter_callback(msg):
...
                         help='rmw_implementation identifier')
-    if args is None:
-        parsargs = parser.parse_args()
-    else:
-        parsargs = parser.parse_args(args)
+    args = parser.parse_args(argv)
 
-    if parsargs is not None:
-        if parsargs.rmw_implementation is not None:
+    if args is not None:
```
dirk-thomas: Obsolete condition.

mikaelarguedas:
moved to https://github.com/ros2/tutorials/pull/3

PR #130, opened 2016-10-27
Updated the service server example

haueck:
There must be at least one shared pointer to the service instance
so it is not destroyed.

### DIFF
--- a/rclcpp_examples/src/services/add_two_ints_server.cpp
+++ b/rclcpp_examples/src/services/add_two_ints_server.cpp
@@ -36,7 +36,7 @@ int main(int argc, char ** argv)
 
   auto node = rclcpp::Node::make_shared("add_two_ints_server");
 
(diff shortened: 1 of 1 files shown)


rclcpp_examples/src/services/add_two_ints_server.cpp
```diff
@@ -36,7 +36,7 @@ int main(int argc, char ** argv)
 
   auto node = rclcpp::Node::make_shared("add_two_ints_server");
 
-  node->create_service<example_interfaces::srv::AddTwoInts>("add_two_ints", handle_add_two_ints);
+  auto client = node->create_service<example_interfaces::srv::AddTwoInts>("add_two_ints", handle_add_two_ints);
```
dirk-thomas: The variable should be called `server` instead of `client`.

haueck: Fixed

wjwwood:
Fixed in e2bb6e364fa5edd972f3775d9f18f546c7e2eca9, verified it passed uncrustify and cpplint locally.

PR #131, opened 2016-10-29
Re-enable service tests for connext

dhood:
This probably could have been changed since https://github.com/ros2/rmw_connext/pull/165

This PR doesn't necessarily connect to https://github.com/ros2/rmw_connext/issues/168, but it has the same branch name so that these tests can be enabled on the CI jobs for the other PRs with the `local_graph_changes` branch name.


dhood:
correction: this could not have been merged with ros2/rmw_connext#165 (because of https://github.com/ros2/rmw_connext/pull/165#issuecomment-222053106)

connects to ros2/rmw_connext#168

PR #132, opened 2016-10-31
[rclpy_examples] QoS examples

mikaelarguedas:
Add reliable/best effort talker listener example

### DIFF
--- a/rclpy_examples/listener_qos_py.py
+++ b/rclpy_examples/listener_qos_py.py
@@ -1,0 +1,47 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+import rclpy
+from rclpy.qos import qos_profile_default, QoSReliabilityPolicy, QoSHistoryPolicy
+
+from std_msgs.msg import String
+
+
+def chatter_callback(msg):
+    print('I heard: [%s]' % msg.data)
+
+
+def main(args=None):
+    if args is None:
+        args = sys.argv
+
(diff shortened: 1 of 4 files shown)


rclpy_examples/listener_qos_py.py
```diff
@@ -0,0 +1,58 @@
...
+
+def chatter_callback(msg):
+    print('I heard: [%s]' % msg.data)
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '-r', '--reliability', type=int, default=0, choices=[0, 1],
+        help='0: reliable, 1: best effort')
```
dirk-thomas: When being called `reliability` I would expect the value `1` to actually be reliable.

mikaelarguedas: thanks for pointing it out, fixed in https://github.com/ros2/examples/pull/132/commits/1ce623944a512c9c2f86a1d0ba6583856f4a2619

rclpy_examples/listener_qos_py.py
```diff
@@ -0,0 +1,58 @@
...
+
+def chatter_callback(msg):
+    print('I heard: [%s]' % msg.data)
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '-r', '--reliability', type=int, default=0, choices=[0, 1],
+        help='1: reliable, 0: best effort')
```
dirk-thomas: Are there any other options planned? If not I would suggest using a boolean flag `--reliable` instead.

dirk-thomas:
Does it make sense to create new scripts rather then extending the already existing ones?

mikaelarguedas:
I see your point, this comes down to how simple do we want the examples to be, should the basic pub/sub examples expose how to configure the Quality of Service? or does it make sense to have another script showing how to handle QoS on top of the basic example?
IMO the code in this PR could be extended to expose more quality of service settings in the future and could be used as a full QoS example/tutorial and should stay dissociated from the minimalist pub/sub.
(this PR also has the goal of giving a starting point for https://github.com/ros2/demos/issues/85).

mikaelarguedas:
thanks

mikaelarguedas:
If nobody has strong opinions about https://github.com/ros2/examples/pull/132#issuecomment-257907246 I'm going to merge this as is

dirk-thomas:
Sure, go ahead with the merge.

PR #135, opened 2016-11-21
standalone simple packages for rclcpp examples

codebot:
The intent is that this directory tree will eventually replace the long-form material in `rclcpp_examples` which has migrated to https://github.com/ros2/tutorials


mikaelarguedas:
@codebot maybe we should sync tutorials on the last state of examples before making any further change to this repo

mikaelarguedas:
@codebot I pushed the [missing commit](https://github.com/ros2/tutorials/commit/b33d3007fd1e6c6622555d7ee7bbcde0b8f00500) on ros/tutorials so we're good to move forward on both repos

PR #137, opened 2016-11-22
add examples_ prefix to package names to avoid future collisions

codebot:
Some of the package names we're coming up with in the `examples` repo are pretty generic. Prefixing them with `examples_` will reduce the likelihood of colliding with other package names.


mikaelarguedas:
should the generated targets be prefixed as well ?

mikaelarguedas:
yeah I agree, we use publisher/subscriber and requester/replier in a lot of places maybe it would make sense to update the vocabulary everywhere.

rclcpp/minimal_subscriber/lambda.cpp
```diff
@@ -2,11 +2,11 @@
...
 #include "std_msgs/msg/string.hpp"
 
-struct Listener : public rclcpp::Node
+struct MinimalSubscriber : public rclcpp::Node
 {
-  Listener() : Node("listener")
+  MinimalSubscriber() : Node("minimal_subscriber")
   {
-    subscription_ = this->create_subscription<std_msgs::msg::String>("chatter",
+    subscription_ = this->create_subscription<std_msgs::msg::String>("topic",
```
mikaelarguedas: nicktip: I think we encourage to wrap after the parenthesis
same below

rclcpp/minimal_publisher/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+add_executable(${PROJECT_NAME}_member_function member_function.cpp)
+ament_target_dependencies(${PROJECT_NAME}_member_function rclcpp std_msgs)
+
+add_executable(${PROJECT_NAME}_not_composable not_composable.cpp)
+ament_target_dependencies(${PROJECT_NAME}_not_composable rclcpp std_msgs)
+
+install(TARGETS ${PROJECT_NAME}_lambda
+                ${PROJECT_NAME}_member_function
+                ${PROJECT_NAME}_not_composable
+        DESTINATION bin)
```
dirk-thomas: Please avoid aligned indentation.

rclcpp/minimal_publisher/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.5)
+project(examples_rclcpp_minimal_publisher)
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(std_msgs REQUIRED)
+if (NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+endif()
```
dirk-thomas: I would suggest moving the C++ flags above the `find_package` (as it is done in all other packages) since the CMake config files could already try to use them.

rclcpp/minimal_publisher/lambda.cpp
```diff
@@ -2,13 +2,14 @@
...
+ * code somewhat more difficult to understand at first glance. */
 
-struct Talker : public rclcpp::Node
+struct MinimalPublisher : public rclcpp::Node
 {
-  Talker() : Node("talker"), count_(0)
+  MinimalPublisher() : Node("minimal_publisher"), count_(0)
   {
-    publisher_ = this->create_publisher<std_msgs::msg::String>("chatter");
+    publisher_ = this->create_publisher<std_msgs::msg::String>("topic");
```
dirk-thomas: Why change the topic name? This will make the code incompatible with the ROS 1 equivalents.

dirk-thomas:
Since the packages (intentionally) don't contain any automatic linting did you run all linters on the example code?

rclcpp/minimal_publisher/lambda.cpp
codebot: That is true, but the ROS 1 equivalent has moved to http://github.com/ros2/tutorials   where it will (eventually) have lots of inline comments and narrative documentation like in the [ROS 1 code](https://github.com/ros/ros_tutorials/blob/kinetic-devel/roscpp_tutorials/talker/talker.cpp). The idea for this repository is to have lots of short copy-paste friendly examples that are as unambiguous as possible. Hence the name change from `talker` to `minimal_publisher`, etc.

rclcpp/minimal_publisher/CMakeLists.txt
codebot: OK. For posterity, I'll add some text to the [style guide](https://github.com/ros2/ros2/wiki/Developer-Guide#cmake) which makes this more clear.

dirk-thomas:
For each linter you will find the way to invoke it in its documentation (e.g. https://github.com/ament/ament_lint/blob/master/ament_pep8/doc/index.rst).

dirk-thomas:
Are additional rclcpp examples planned for a future PR? The ones in this PR look fine to me.

dirk-thomas:
Maybe we should also have examples how we recommend users to write their nodes in the future, inheriting from node and creating a library.

dirk-thomas:
Please squash your commits before / during the merge.

wjwwood:
Technically no, but we shouldn't make a habit of force pushing to master. I'd say just leave it and try to remember in the future 😄.

PR #138, opened 2016-11-23
Minimal service and client

codebot:
builds on directory reorganization of #137 

adds minimal examples of service and client packages


mikaelarguedas:
Do you plan on providing the same set of strategies as for the pub sub in these packages (lambda, member_functions etc)?

rclcpp/minimal_service/not_composable.cpp
```diff
@@ -0,0 +1,38 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include "rclcpp/rclcpp.hpp"
+#include "examples_rclcpp_minimal_service/srv/add_two_ints.hpp"
+
+using Service = examples_rclcpp_minimal_service::srv::AddTwoInts;
```
wjwwood: I'd recommend two things here:

- use `example_interfaces/srv/AddTwoInts`
- do `using example_interfaces::srv::AddTwoInts;` and replace `Service` with `AddTwoInts` throughout

I don't know if there was a reason for creating the service locally, but every message and/or service we have to build on the farm is lots of extra time. As for illustrating how to make your own service, perhaps we could just leave the machinery for that commented out or just refer to the `example_interfaces` package, it is quite simple already:

https://github.com/ros2/examples/blob/master/example_interfaces/CMakeLists.txt

I'm less passionate about replacing `Service` with `AddTwoInts`, but I think it does give more meaning to each place it is used (in the callback and at the creation of the service). It gets really important if the user is going to create two services in the same file.

rclcpp/minimal_client/not_composable.cpp
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include "rclcpp/rclcpp.hpp"
+#include "examples_rclcpp_minimal_service/srv/add_two_ints.hpp"
+
+using Service = examples_rclcpp_minimal_service::srv::AddTwoInts;
+using Request = Service::Request;
```
wjwwood: Same comment as in the other file, I get what you're trying to do here with shorter lines and simpler names, but I think `AddTwoInts::Request` is so much more informative than `Request` or `Service::Request`, which require me to look up the translation.

rclcpp/minimal_client/not_composable.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  auto node = rclcpp::node::Node::make_shared("minimal_subscriber");
```
wjwwood: This looks like a copy-paste error, right?

rclcpp/minimal_service/not_composable.cpp
codebot: I created the service locally on purpose here, to show the CMake pattern for doing so (since it's somewhat non-trivial to ensure the service is generated first before Make tries to build the code which uses it), but it's a fair point that this is no longer "minimal." I'll take your suggestion now and use `example_interfaces`, and in the future, if anyone cares, we can create a package which both generates its own service and uses it.

rclcpp/minimal_client/not_composable.cpp
codebot: doh :disappointed:

wjwwood:
lgtm, thanks!

dirk-thomas:
@codebot It looks like this PR has introduced new warning on the farm (http://ci.ros2.org/job/ci_osx/1554/warnings10Result/ http://ci.ros2.org/job/ci_windows/2006/warnings41Result/). Please address them.

PR #139, opened 2016-11-28
Rclpy minimal pub sub

mikaelarguedas:
This PR adds several simple pub/sub examples.
This not sure if the "local function" and the "lambda" are necessary, maybe we should just pick one and discard the other.


rclpy/topics/minimal_publisher/README.md
```diff
@@ -0,0 +1,6 @@
+# Minimal "publisher" cookbook recipes
+
+This package contains a few different strategies for creating short nodes that blast out messages.
+The `publisher_old_school` recipe creates a talker node very similar to how it would be done in ROS1 using rospy.
```
dhood: can you use ROS 1 and ROS 2 (with spaces) for consistency please

rclpy/topics/minimal_publisher/README.md
```diff
@@ -0,0 +1,6 @@
...
+The `publisher_lambda` and `publisher_local_function` allow to leverage the timers provided by ROS2 to trigger message publication.
+The `publisher_member_function` created a class MinimalPublisher that sends messages periodically
```
codebot: I 'member!

rclpy/topics/minimal_publisher/publisher_lambda.py
```diff
@@ -0,0 +1,47 @@
...
+    timer = node.create_timer(0.5, lambda: publisher_.publish(msg))
+
+    while rclpy.ok():
+        msg.data = 'Hello World: %d' % i
+        i += 1
+        print('Publishing: "%s"' % msg.data)
+        publisher_.publish(msg)
+        rclpy.spin_once(node)
+
+    node.destroy_timer(timer)
```
codebot: do these timers and nodes get destroyed automatically on shutdown, or is manual destruction necessary?

rclpy/topics/minimal_publisher/publisher_lambda.py
```diff
@@ -0,0 +1,47 @@
...
+
+    msg = String()
+    i = 0
+    timer = node.create_timer(0.5, lambda: publisher_.publish(msg))
+
+    while rclpy.ok():
+        msg.data = 'Hello World: %d' % i
+        i += 1
+        print('Publishing: "%s"' % msg.data)
+        publisher_.publish(msg)
```
codebot: I'm probably just not understanding this correctly, but it seems like the timer with lambda will also be publishing, so the messages will be sent twice (one from the lambda, one from this `publish()` call ? And will this loop run at a particular rate, or will it run as quickly as possible?

rclpy/topics/minimal_publisher/publisher_local_function.py
```diff
@@ -0,0 +1,51 @@
...
+        nonlocal i
+        msg.data = 'Hello World: %d' % i
+        i += 1
+        print('Publishing: "%s"' % msg.data)
+        publisher_.publish(msg)
+
+    timer = node.create_timer(0.5, timer_callback)
+
+    while rclpy.ok():
+        rclpy.spin_once(node)
```
codebot: Is it recommended to have this `while` loop calling `spin_once()` or is it better to just have `spin()`  (I don't know) :chicken:

rclpy/topics/minimal_publisher/publisher_old_school.py
```diff
@@ -0,0 +1,48 @@
...
+
+    msg = String()
+
+    i = 0
+    while rclpy.ok():
+        msg.data = 'Hello World: %d' % i
+        i += 1
+        print('Publishing: "%s"' % msg.data)
+        publisher_.publish(msg)
+        # TODO(mikaelarguedas): explain why spin_once can't be used here
```
dhood: is it possible to add a comment to address this now?

rclpy/topics/minimal_publisher/publisher_lambda.py
mikaelarguedas: they do get destroyed automatically, I added the destruction to avoid that linters complain about unused variables. I'm unsure about the amount of comments that should be placed in these examples. But this one definitely deserves deserves one.

rclpy/topics/minimal_publisher/package.xml
```diff
@@ -0,0 +1,22 @@
...
+  <maintainer email="mikael@osrfoundation.org">Mikael Arguedas</maintainer>
+  <license>Apache License 2.0</license>
+
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+
+  <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_pep257</test_depend>
+  <test_depend>ament_pep8</test_depend>
+  <test_depend>ament_pyflakes</test_depend>
```
dhood: do you think the linters go against the minimalistic nature? there seems to be a few lines added that could confuse people just for linter sake

rclpy/topics/minimal_publisher/publisher_member_function.py
```diff
@@ -0,0 +1,57 @@
...
+
+import rclpy
+
+from std_msgs.msg import String
+
+
+class MinimalPublisher:
+
+    def __init__(self):
+        self.node = rclpy.create_node('minimal_publisher')
```
codebot: I'm not sure about the python recommended style, but having the lines smashed together without whitespace seems nicer to me. But maybe there is a style guide for such things :chicken:

rclpy/topics/minimal_subscriber/subscriber_lambda.py
```diff
@@ -0,0 +1,41 @@
...
+import rclpy
+
+from std_msgs.msg import String
+
+
+def main(args=None):
+    if args is None:
+        args = sys.argv
+
+    rclpy.init(args)
```
dhood: It would be more minimalistic to remove the `args` variable altogether (and `sys` import)

rclpy/topics/minimal_subscriber/subscriber_local_function.py
```diff
@@ -0,0 +1,44 @@
...
+    rclpy.init(args)
+
+    node = rclpy.create_node('minimal_publisher')
+
+    def listener_callback(msg):
+        print('I heard: [%s]' % msg.data)
+
+    subscription_ = node.create_subscription(
+        String, 'topic', listener_callback)
+    subscription_  # noqa
```
codebot: what does this line do? ELI5

rclpy/topics/minimal_publisher/publisher_lambda.py
```diff
@@ -0,0 +1,47 @@
...
+        args = sys.argv
+
+    rclpy.init(args)
+
+    node = rclpy.create_node('minimal_publisher')
+    publisher_ = node.create_publisher(String, 'topic')
+
+    msg = String()
+    i = 0
+    timer = node.create_timer(0.5, lambda: publisher_.publish(msg))
```
dhood: I'd be in favour of removing/reworking this script since the lambda doesn't show that much (like you mentioned)

rclpy/topics/minimal_publisher/publisher_lambda.py
mikaelarguedas: done in https://github.com/ros2/examples/pull/139/commits/a55f5b1a25d2d7424a101e1b2e66754d4e7bf377

rclpy/topics/minimal_subscriber/subscriber_member_function.py
```diff
@@ -0,0 +1,53 @@
...
+    rclpy.init(args)
+
+    minimal_subscriber = MinimalSubscriber()
+    while rclpy.ok():
+        rclpy.spin_once(minimal_subscriber.node)
+
+    # Destroy the node explicitly
+    # (optional - otherwise it will be done automatically
+    # when the garbage collector destroys the node object)
+    minimal_subscriber.node.destroy_node()
```
codebot: can this be done in some sort of destructor for the `MinimalSubscriber` class?

rclpy/topics/minimal_publisher/publisher_local_function.py
mikaelarguedas: unfortunately there is no `spin` yet in python, only `spin_once`

rclpy/topics/minimal_publisher/publisher_lambda.py
mikaelarguedas: >_< indeed this publish shouldn't be here.
Regarding the loop, filling up the message etc cannot be done in the lambda because of how lambdas work in python (no variable assignment, has to be single line etc). I would be in favor of removing the `publisher_lambda` and the `subscriber_local_function` but I'm not a fan of not having symmetry in the minimalist publisher/subscriber examples

rclpy/topics/minimal_publisher/package.xml
mikaelarguedas: I though the conclusion was that examples should be tested and linted as opposed to the tutorials. Did I miss something?

rclpy/topics/minimal_subscriber/subscriber_local_function.py
mikaelarguedas: This line tells the linter that this variable is not an unused variable. That's the counterpart of calling the destroy_* functions explicitely.

rclpy/topics/minimal_publisher/package.xml
wjwwood: I thought the conclusion was exactly the opposite, because testing would cause the CMake and potentially the code to be more complex, making it less copy-pastable. The tutorials, however, are not meant as minimal starting locations, so they can afford to have more complex CMake lists and code?

dhood: tutorial code tested, examples copy-paste friendly

codebot: Hmm I thought it was the other way around (examples are as short as possible, but tutorials are longer-form with narrative and automatically tested)? :chicken:

mikaelarguedas: Ok my bad I thought we wanted to make sure that the copy-pasted code compiles/run whereas it was not crucial for the ones supposed to show concepts on the wiki. Will remove all the tests then

wjwwood: We do want to make sure it compiles, and I don't think that linting it is a bad idea. If our code commonly make the sure do something that doesn't pass a linter, then we should consider changing the API.

So, I'd say leave the linters, but do not test it with automated tests.

mikaelarguedas: I left the linters and added a comment in the package.xml to highlight the purpose of these dependencies

rclpy/topics/minimal_publisher/publisher_old_school.py
mikaelarguedas: that has been done in https://github.com/ros2/examples/pull/139/commits/1cf6702883d0e2553562ca79c3e63ffd6e7072a4

rclpy/topics/minimal_subscriber/subscriber_local_function.py
mikaelarguedas: Comment change from to avoid confusion

mikaelarguedas:
Comments addressed

rclpy/topics/minimal_publisher/README.md
```diff
@@ -2,5 +2,5 @@
 
 This package contains a few different strategies for creating short nodes that blast out messages.
 The `publisher_old_school` recipe creates a talker node very similar to how it would be done in ROS 1 using rospy.
-The `publisher_local_function` allow to leverage the timers provided by ROS 2 to trigger message publication.
-The `publisher_member_function` created a class MinimalPublisher that sends messages periodically
+The `publisher_local_function` recipe shows how to leverage the timers provided by ROS 2 to trigger message publication.
+The `publisher_member_function` recipe creates a class MinimalPublisher that sends messages periodically.
```
mikaelarguedas: thx

mikaelarguedas:
Thanks everyone for reviewing.
Any other comments on this ?

rclpy/topics/minimal_publisher/publisher_member_function.py
```diff
@@ -0,0 +1,53 @@
...
+        self.node.destroy_node()
+
+
+def main(args=None):
+    rclpy.init(args)
+
+    minimal_publisher = MinimalPublisher()
+
+    while rclpy.ok():
+        rclpy.spin_once(minimal_publisher.node)
```
dhood: since the cookbook examples are supposed to show best practice it might be best to not have the node attached to the class, so that multiple publishers can easily be attached to a node. it's not a blocker though, just a suggestion

mikaelarguedas: good point, done https://github.com/ros2/examples/pull/139/commits/eea27f3cbb2aaceedfedf5d7f48b6b4ce3124b75

rclpy/topics/minimal_publisher/publisher_local_function.py
```diff
@@ -0,0 +1,49 @@
...
+    i = 0
+
+    def timer_callback():
+        nonlocal i
+        msg.data = 'Hello World: %d' % i
+        i += 1
+        print('Publishing: "%s"' % msg.data)
+        publisher_.publish(msg)
+
+    timer = node.create_timer(0.5, timer_callback)
```
dhood: I think it's worth highlighting the units here, since they're different to rclcpp. either commented or through a variable name

mikaelarguedas: I don't think a comment is necessary given that we pass the same value to the python `sleep` function. It should be pretty obvious given that all scripts have equivalent behavior. If doubt there is, the variable name in the API is [timer_period_sec](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/node.py#L96).
I'm not opposed to it but don't think this is a source of confusion

rclpy/topics/minimal_publisher/test/test_copyright.py
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ament_copyright.main import main
+
+
+def test_copyright():
+    rc = main(argv=['.', 'test'])
```
dirk-thomas: Why does this need to be `.` and `test`? Isn't `test` a subfolder of the package root `.`?

rclpy/topics/minimal_publisher/publisher_local_function.py
dhood: ok I added a comment in https://github.com/ros2/examples/pull/139/commits/c056f8f70c2c4200c6e1521ecdc9289cf5a4871b just in case

PR #141, opened 2016-11-28
templates pub sub

mikaelarguedas:
This PR provides message agnostic publisher/subscriber that could be used as a base for rostopic pub/echo implementations.
Looks like cookbook material to me but I'd like the team's opinion on it to be sure that's where this belongs

### DIFF
--- a/rclpy/topics/generic_publisher/generic_publisher.py
+++ b/rclpy/topics/generic_publisher/generic_publisher.py
@@ -1,0 +1,56 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


mikaelarguedas:
There is a way to stuff the message from the command line, it would require using the ament_index to find all the messages definitions available in the workspace and then create a string in json format that people will stuff with their values. Then on the python side it should be pretty straight forward to get the dictionnary, populate and send the message.
What we are lacking for now is the remote topics discovery which is why I don't expose a topic argument in this script.
To summary we can:
- add the message stuffing ourselves and postpone exposing/autocompleting the topic name until we have the underlying machinery
- merge the current feature set as a skeleton of what `rostopic pub/echo` would look like and open the message stuffing + topic names implementation to the community

Good point, I'll had an usage string

Regarding the "Ctrl+C" being ineffective, that is a known issue, there are 2 ways around it:
- if we have time, implement `guard_conditions` in Python to be able to interrupt the C spin function
- change the default timeout of spin to something else than "wait forever" so that Python can process the interrupt and kill the script
We will definitely need to do one of the two before the beta because all subscription/service servers scripts cannot be killed properly with the current state

mikaelarguedas:
The help now displays the default values of all arguments
I added the possibility of providing values to the publisher in the form of a dictionary in json format. I'm going to pause this until we have a clear idea of how we expect user to provide values for message population (ros1 rostopic pub positional argument approach or a string in a specific format).

mikaelarguedas:
closed in favor of https://github.com/ros2/cli_tools/issues/3

PR #143, opened 2016-12-01
use new rclcpp::literals namespace

Sarcasm:
Related to ros2/rclcpp#284.

Connects to ros2/rclcpp#282

### DIFF
--- a/rclcpp/minimal_publisher/lambda.cpp
+++ b/rclcpp/minimal_publisher/lambda.cpp
@@ -15,6 +15,8 @@
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/string.hpp"
 
+using namespace rclcpp::literals;
+
 /* This example creates a subclass of Node and uses a fancy C++11 lambda
  * function to shorten the callback syntax, at the expense of making the
  * code somewhat more difficult to understand at first glance. */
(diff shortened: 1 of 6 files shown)


wjwwood:
@codebot sorry but this is going to bite into your minimalism by two more lines 😨.

wjwwood:
This will go away again (or at least change) if we switch to using the `std::chrono` literals with C++14, which we're discussing internally right now.

Sarcasm:
For some of the examples, the fix is easy.

Don't use the literals:

```cpp
using namespace rclcpp::literals;
timer_ = this->create_wall_timer(500_ms, timer_callback);
```

becomes

```cpp
timer_ = this->create_wall_timer(std::chrono::milliseconds(500), timer_callback);
```

It looks good enough.


If the examples were mixing different units like the example below,
then it would be unfortunate,
but if it's for one value I think the non-user-defined literal code is quite fine.

```cpp
timer_ = this->create_wall_timer(250_s + 500_ms + 3_ns, timer_callback);
```

wjwwood:
I'm going to close this one, because it was completely covered by @codebot's C++14 changes and/or packages that were moved (deleted from this repository). I'll make sure to touch the new packages added in https://github.com/ros2/demos/pull/94 one way or the other.

PR #144, opened 2016-12-01
Example of running node in another thread

dhood:
This adds an example of plotting in a main thread and processing subscription callbacks in a different thread.

you get the following output (note the plot update doesn't block the 10Hz data callback)

```
Data received: -1
Data received: 1
Data received: -1
Updating plot
Data received: 1
Data received: -1
Data received: 1
Data received: -1
Data received: 1
Data received: -1
Data received: 1
Data received: -1
Data received: 1
Finished updating plot
Data received: -1
Data received: 1
```


dhood:
thanks go to @mikaelarguedas for helping me get the threading working the other day!

rclpy/threading/data_plotter.py
```diff
@@ -0,0 +1,117 @@
...
+        plt.ion()
+        self.fig = plt.figure()
+        self.received_data = []
+        # As the received data can be access by both the main plotting thread and the thread
+        # processing data callbacks, its access must be protected by a lock
+        self.received_data_lock = Lock()
+
+    # Called when new data is received
+    def data_callback(self, msg):
+        print('Data received: {0}'.format(msg.data))
```
wjwwood: I personally prefer the `.format()`, but either is fine. I'm not sure if we ever decided to only do one or the other.

rclpy/threading/data_plotter.py
```diff
@@ -0,0 +1,117 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from copy import deepcopy
+from threading import Lock, Thread
+import time
+
+import rclpy
+
+import matplotlib.pyplot as plt
```
wjwwood: Also, we might want to express a dependency on `matplotlib` in the `package.xml` or similar, even though we don't use that yet to install dependencies, a la `rosdep`.

rclpy/threading/data_plotter.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
wjwwood: It's true that won't run as you'd expect on Windows, the only portable way to do that is with a setuptools entry point, so that it creates an `.exe` file for the script on Windows. I think it's fine to leave it though, and Windows users will need to do `python3 path/to/data_plotter.py` instead of `./path/to/data_plotter.py`.

It should be fine, so long as we're not running this in an automated way on Windows.

rclpy/threading/data_plotter.py
dhood: like this yeah? https://github.com/ros2/examples/pull/144/commits/88ec93bb956bb8358e29ae3db31269e42817ab17 or something else

mikaelarguedas:
Looks nice, a few nicktips and may require linting / testing on all OS but the code looks good to me

rclpy/threading/data_plotter.py
mikaelarguedas: Yeah both could work, I would prefer being consistent across basic examples though.
I think that if you source/call `local_setup.bat` you can just run the executable that is on your path exactly like in linux, isn't it ?

rclpy/threading/setup.py
```diff
@@ -0,0 +1,30 @@
...
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python',
+        'Topic :: Software Development',
+    ],
+    description='Package containing examples of using threading with rclpy.',
+    license='Apache License, Version 2.0',
+    entry_points={
+        'console_scripts': [
+            'data_plotter = data_plotter:main',
+            'data_publisher = data_publisher:main',
```
dhood: thanks for catching that, the names are quite generic indeed

rclpy/threading/data_plotter.py
wjwwood: That is true.

rclpy/threading/data_plotter.py
mikaelarguedas: not sure what the right way to declare third party deps is :confused: . I would expect that these packages will be installed via pip on other platforms where the module name will be `matplotlib` and not `python3_matplotlib`.

dhood: yeah, I know what you mean - I was copying https://github.com/ament/ament_lint/blob/853968a9befe153e8c7d71659bbdceb18c97d9a2/ament_pep8/package.xml#L13 but am happy to adapt

wjwwood: Just use the roadworks entry for matplotlib for now: https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L1061

wjwwood: Err entry not roadworks?

wjwwood: Sorry, `rosdep` not `roadworks`... Autocorrect fail.

wjwwood: Oh well, I guess the `python3-matplotlib` might make more sense... I'm not sure. In the past we have had `trusty` and `trusty_py3` as different "operating systems" to handle the python and python3 issues with dependencies. Honestly, something or anything in the `package.xml` which mentions we use `matplotlib` here is fine.

wjwwood: 88ec93bb956bb8358e29ae3db31269e42817ab17 lgtm, sorry for the spam.

rclpy/threading/package.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>examples_rclpy_threading</name>
+  <version>0.0.0</version>
+  <description>Package with examples of using threading with rclpy.</description>
+  <maintainer email="dhood@osrfoundation.org">D. Hood</maintainer>
+  <license>Apache License 2.0</license>
+
+  <exec_depend>python3-matplotlib</exec_depend>
```
dirk-thomas: Just as a comment: this rosdep key doesn't exist at the moment.

rclpy/threading/data_plotter.py
```diff
@@ -0,0 +1,114 @@
...
+        # processing data callbacks, its access must be protected by a lock
+        self.received_data_lock = Lock()
+
+    def data_callback(self, msg):
+        print('Data received: {0}'.format(msg.data))
+
+        # Store received data
+        self.received_data_lock.acquire()
+        self.received_data.append(msg.data)
+        self.received_data_lock.release()
```
dhood: thank you for this and other tips! I made the changes in https://github.com/ros2/examples/pull/144/commits/71db410c064a14d8e47ae14e338fa0459ffb90bc

rclpy/threading/package.xml
dhood: I have made a note so that this is clear to others looking at the code in https://github.com/ros2/examples/pull/144/commits/553ca84bbed9e1fea36cb5ccdf734af226d3f111

dhood:
thanks for the reviews - this is a spinoff from https://github.com/ros2/demos/pull/90 so I propagated the comments into that PR.

This needed some updates since https://github.com/ros2/rclpy/pull/61 but should be good now. tested it on windows manually

mikaelarguedas:
lgtm, thanks for iterating on it

dirk-thomas:
This failed a test on Windows: http://ci.ros2.org/job/ci_windows/lastCompletedBuild/testReport/nose.failure/Failure/runTest/

dhood:
I'll make it so this doesn't cause a test failure but I actually don't see why nosetests is being called on this package because there are no tests specified in the setup.py https://github.com/ros2/examples/blob/318d22cc440bb20a241ec455344ec2d73a0cab3f/rclpy/threading/setup.py

dirk-thomas:
Likely `nosetests` will import all your files for the coverage report. The nodes probably need the dependency to be installed.

dhood:
thanks for the explanation. I'll revert this until I get it to pass coverage tests without matplotlib installed (since it shouldn't be a requirement for users)

PR #150, opened 2016-12-12
remove unused LargeFixed msg

mikaelarguedas:
not used since https://github.com/ros2/examples/pull/67


example_interfaces/CMakeLists.txt
```diff
@@ -11,9 +11,7 @@ find_package(builtin_interfaces REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 
 rosidl_generate_interfaces(${PROJECT_NAME}
-  "msg/LargeFixed.msg"
   "srv/AddTwoInts.srv"
-  DEPENDENCIES builtin_interfaces
```
dirk-thomas: I `builtin_interfaces` is not used anymore anywhere it should also be removed from the manifest.

mikaelarguedas: good point https://github.com/ros2/examples/pull/150/commits/f1c5bddb22721f87565cf159a59b860ba971327d

PR #156, opened 2016-12-15
Comment mapping_rules.yaml

dhood:
its purpose is not clear when looking at the example_interfaces package alone


dirk-thomas:
The purpose is clear when looking at the package manifest. The file is being referenced using the tag `ros1_bridge`.

The comment in the yaml is benefitiial nevertheless. :+1:

dhood:
yep, good point. merging thanks.

PR #159, opened 2016-12-18
print hello world from timer callbacks

codebot:
as far as I understand, this is the simplest possible example of a wall timer.

### DIFF
--- a/rclcpp/minimal_timer/lambda.cpp
+++ b/rclcpp/minimal_timer/lambda.cpp
@@ -1,0 +1,43 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


rclcpp/minimal_timer/lambda.cpp
```diff
@@ -0,0 +1,43 @@
...
+ * code somewhat more difficult to understand at first glance if you are
+ * unaccustomed to */
+
+class MinimalTimer : public rclcpp::Node
+{
+public:
+  MinimalTimer()
+  : Node("minimal_timer")
+  {
+    auto timer_callback = [this]() -> void { printf("Hello, world!\n"); };
```
wjwwood: You don't need to capture `this` I think.

codebot: aha! indeed, you're right. I simplified it. I'll kick off a CI now just to be sure it's fine on all platforms.

PR #163, opened 2017-01-04
Revert "readd dummy message"

mikaelarguedas:
Reverts ros2/examples#155
Depends on https://github.com/ros2/rosidl/pull/198
Connects to https://github.com/ros2/rosidl/issues/193


mikaelarguedas:
This message has been added only to work around a typesupport issue when using packages that contain only services and no messages. So this message is only a placeholder not used by anyone.
https://github.com/ros2/rosidl/pull/198 fixes the missing symbols so this placeholder won't be necessary once https://github.com/ros2/rosidl/pull/198 merged

PR #165, opened 2017-01-04
create missing __init__.py files

mikaelarguedas:
`setuptools` print a message to stderr if `__init__.py` don't exist for each pymodule listed in setup.py
Search for "package init file" in [the last nightly](http://ci.ros2.org/view/nightly/job/nightly_linux_debug/320/consoleFull)


dirk-thomas:
These files seem to be outside of the package (parent folder of where the `package.xml` files are). That doesn't look right.

mikaelarguedas:
Fixed, thanks

wjwwood:
I don't think this is the best way to store and distrubute the "Python modules" or "Python package". Our package layout guide recommends that you put this code in a folder `src/<package name>` and that scripts are placed into the `script` folder. I think this is a pretty useway to layout the code. I think it is non-standard to have the setup.py in the same folder as the `__init__.py`, the setuptools convention is to have the source code in a folder called `<project name>`. For example:

http://python-packaging.readthedocs.io/en/latest/minimal.html#creating-the-scaffolding

dirk-thomas:
It is also fine to skip the `__init__py` file if the intend is to allow spreading the portions over multiple locations rather then a regular package in a single location (see [PEP420](https://www.python.org/dev/peps/pep-0420/)).

mikaelarguedas:
yeah that was the trade off of these examples. Having a flat tree where users can see all the files at the root of the package folder (like in the c++ examples) vs show good practice and create the expected folder structure.

Regarding skipping the `__init__.py` files I'm ok with it. I created this PR just because I was trying to get rid of all messages printed to stderr. The examples work just as well without them

dirk-thomas:
Closing for now.

PR #166, opened 2017-02-01
remove unused srv file

mikaelarguedas:
this example uses the service defined in [example_interfaces](https://github.com/ros2/examples/blob/25c091cf4418f22437fe61dcf18753310a501e03/example_interfaces/srv/AddTwoInts.srv) package so this one can be removed


dhood:
we can remove this too yeah? https://github.com/ros2/examples/blob/6b412f2c166e926da3fd6b18b937708564153873/rclcpp/minimal_service/package.xml#L11

mikaelarguedas:
indeed! thanks for pointing it out

PR #177, opened 2017-06-28
rename executables with shorter names

### LINKED ISSUE #174: Remove package prefix from entry point / executable names
Now that the executables are installed in libexec all executable names should be simplified:
e.g. `examples_rclpy_minimal_publisher_member_function` -> `minimal_publisher_member_function`
or 
`publisher_member_function`

mikaelarguedas:
NOTE: I changed only the names of the executables and not the package names

fixes #174 
connects to #174


wjwwood:
I don't think these are referenced from the wiki, but it might be worth a couple greps through the wiki's git clone to spot check it.

mikaelarguedas:
Didnt find any reference of these on the wiki. Merging this

PR #178, opened 2017-08-09
remove dependency on ament_python and perform customizations in setup.py

dirk-thomas:
Connect to ament/ament_python#3.


rclpy/services/minimal_client/setup.cfg
```diff
@@ -0,0 +1,4 @@
+[develop]
+script-dir=$base/lib/examples_rclpy_minimal_client
```
dirk-thomas: No, not as a placeholder in part of a string.

PR #183, opened 2017-09-11
remove Listener from the "ThrottledTalkerListener" name given that this is only a throttled talker

mikaelarguedas:


mikaelarguedas:
case sensitive search :/, fixed in https://github.com/ros2/examples/pull/183/commits/8032549a27225bc100673e1a889d5b4038a9894f

PR #184, opened 2017-09-15
Follow up to executor example comments

sloretz:
Follow up to comments on #182 that didn't get addressed before r2b3

* Simplified custom_executor example
* Shortened test paths

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3255/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/546/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2584/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3304/)

### DIFF
--- a/rclpy/executors/examples_rclpy_executors/custom_executor.py
+++ b/rclpy/executors/examples_rclpy_executors/custom_executor.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import threading
+from concurrent.futures import ThreadPoolExecutor
 
 from examples_rclpy_executors.listener import Listener
 from examples_rclpy_executors.talker import Talker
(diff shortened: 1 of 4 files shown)


rclpy/executors/examples_rclpy_executors/custom_executor.py
```diff
@@ -45,16 +45,14 @@ class PriorityExecutor(Executor):
     def __init__(self):
         super().__init__()
         self.high_priority_nodes = set()
-        self.low_priority_thread = None
+        self.hp_executor = ThreadPoolExecutor(max_workers=10)
```
dirk-thomas: I know it is only an example... but `10` seems a little high. Maybe `cpu_count()`?

sloretz: 018513a66e1f025c26ea76112b91f0122090498b

rclpy/executors/examples_rclpy_executors/custom_executor.py
```diff
@@ -45,16 +46,14 @@ class PriorityExecutor(Executor):
     def __init__(self):
         super().__init__()
         self.high_priority_nodes = set()
-        self.low_priority_thread = None
+        self.hp_executor = ThreadPoolExecutor(max_workers=multiprocessing.cpu_count())
```
dirk-thomas: `cpu_count()` may raise `NotImplementedError` which you need to catch and fall back to a hard coded value.

sloretz: How about omitting max_workers? [In python 3.5 it will use os.cpu_count() * 5 if it's implemented, or 5 workers if it's not.](https://github.com/python/cpython/blob/0d68d6d9c6386e3f78893f0506f2ce16d88b5db8/Lib/concurrent/futures/thread.py#L91-L94)

dirk-thomas: Sounds like high defaults when not doing IO-intense operations. Since this example doesn't "do" much it probable doesn't matter. Not sure if it sets a good example - maybe with a comment to document what the default is that is ok.

sloretz: cpu_count() or 4 workers in 5faf21687bb49ee6889b8bf48de6d38532456ee5

rclpy/executors/examples_rclpy_executors/custom_executor.py
```diff
@@ -46,7 +46,7 @@ class PriorityExecutor(Executor):
     def __init__(self):
         super().__init__()
         self.high_priority_nodes = set()
-        self.hp_executor = ThreadPoolExecutor(max_workers=multiprocessing.cpu_count())
+        self.hp_executor = ThreadPoolExecutor(max_workers=(os.cpu_count() or 4))
```
dirk-thomas: Maybe remove the unnecessary parenthesis?

sloretz: 53bd00f6a2a9a6ddb4961ae31d0a138d1bc52354

sloretz:
Builds passed and approved, squash/merging

PR #185, opened 2017-10-13
Use wait_for_service

sloretz:
Replace sleep calls with calls to wait_for_service(). Labeled in-progress because it requires ros2/rclpy#127 to be merged first.

connects to ros2/rclpy#58

### DIFF
--- a/rclpy/services/minimal_client/client.py
+++ b/rclpy/services/minimal_client/client.py
@@ -26,7 +26,7 @@ def main(args=None):
     req.a = 41
     req.b = 1
     while not cli.wait_for_service(timeout_sec=1.0):
-        print("Still waiting for service")
+        print("service not available, waiting again...")
 
     cli.call(req)
     # when calling wait for future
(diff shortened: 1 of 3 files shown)


rclpy/services/minimal_client/client.py
```diff
@@ -27,10 +25,7 @@ def main(args=None):
     req = AddTwoInts.Request()
     req.a = 41
     req.b = 1
-    # TODO(mikaelarguedas) remove this once wait for service implemented
-    # wait for connection to be established
-    # (no wait for service in Python yet)
-    time.sleep(1)
+    cli.wait_for_service()
```
wjwwood: Now that I've approved it 😄, I noticed that this will block forever. If you look at the C++ version it will do wait_for_service for say a second, then loop with a message like "still waiting for service..."

Not required, just an idea to exercise the timeout and return value.

sloretz: af6a2ef9acd03ef48f1c18f2e60ba8c9827329f3

sloretz:
Rebased onto master. I think this can be merged now that ros2/rclpy#161 has been merged. I'll merge it before the end of the day to give time for objections.

PR #186, opened 2017-10-20
Prototype use of rclpy logging for callback_group

dhood:
Is this what we expect this example to look like using rclpy logging?
The script passes its logger (`ros.callback_group_example`) through to the nodes so that the node loggers get attached to the parent logger.

The logger name of the nodes gets the parent logger prepended:
```
[INFO] [ros.callback_group_example.double_talker]: Publishing: "Hello World: 0"
[INFO] [ros.callback_group_example.listener]: I heard: [Hello World: 0]
[INFO] [ros.callback_group_example.double_talker]: Publishing: "Hello World: 1"
[INFO] [ros.callback_group_example.double_talker]: Publishing: "Hello World: 2"
[INFO] [ros.callback_group_example.listener]: I heard: [Hello World: 1]
```

And if you enable debug on the parent logger (by uncommenting https://github.com/ros2/examples/compare/master...use_logging#diff-41999062ba992334cbe0288b618f2a2dR23), the node loggers get debug enabled too:
```
[DEBUG] [ros.callback_group_example.double_talker]: Initialization complete
[DEBUG] [ros.callback_group_example]: Nodes added to executor
[INFO] [ros.callback_group_example.double_talker]: Publishing: "Hello World: 0"
[INFO] [ros.callback_group_example.listener]: I heard: [Hello World: 0]
```

If you don't pass through the script's logger, the nodes will just use independent loggers (still configurable, but not attached to the parent, and the logger name will collide with any other nodes with the same node name).

### DIFF
--- a/rclpy/executors/examples_rclpy_executors/callback_group.py
+++ b/rclpy/executors/examples_rclpy_executors/callback_group.py
@@ -16,14 +16,18 @@ from examples_rclpy_executors.listener import Listener
 import rclpy
 from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
 from rclpy.executors import MultiThreadedExecutor
+import rclpy.logging
 from std_msgs.msg import String
+
+logger = rclpy.logging.get_named_logger('callback_group_example')
+# logger.set_severity_threshold(rclpy.logging.LoggingSeverity.DEBUG)
 
 
 class DoubleTalker(rclpy.Node):
     """Publish messages to a topic using two publishers at different rates."""
 
-    def __init__(self):
-        super().__init__('double_talker')
+    def __init__(self, parent_logger=None):
+        super().__init__('double_talker', parent_logger=parent_logger)
 
         self.i = 0
         self.pub = self.create_publisher(String, 'chatter')
@@ -33,12 +37,13 @@ class DoubleTalker(rclpy.Node):
         # Pass the group as a parameter to give it control over the execution of the timer callback
         self.timer = self.create_timer(1.0, self.timer_callback, callback_group=self.group)
         self.timer2 = self.create_timer(0.5, self.timer_callback, callback_group=self.group)
+        self.logger.debug('Initialization complete')
 
     def timer_callback(self):
         msg = String()
         msg.data = 'Hello World: {0}'.format(self.i)
         self.i += 1
-        print('Publishing: "{0}"'.format(msg.data))
+        self.logger.info('Publishing: "{0}"'.format(msg.data))
         self.pub.publish(msg)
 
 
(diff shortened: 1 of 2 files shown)


dhood:
update: disregard the `ros.` prefix on the logger, that's a separate question.

This is "in review" for discussion but not being proposed for merge (implementation pending decision)

dirk-thomas:
I think it is good that the user can choose to do what is proposed in this patch.

In general I would rather recommend to not embed information about *where* a node is being run into its logger name. E.g. the node might later be stopped dynamically and being restarted in a separate process. From a user point of view I wouldn't expect the logger name of the node to change.

dhood:
If we want to update it to show best practices to users (taking into account what you pointed out about the logger changing if the node is loaded in a different way), what would we recommend? Not passing through the logger for e.g. the listener to attach to, but instead setting a more specific logger name for the node to use? (that specific name would be re-used if we decide to restart it in a separate process)

rclpy/executors/examples_rclpy_executors/callback_group.py
```diff
@@ -16,14 +16,18 @@
 import rclpy
 from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
 from rclpy.executors import MultiThreadedExecutor
+import rclpy.logging
```
dhood: Does it make sense to require users to import the logging module to use it, or should we always import it in rclpy's init instead?

dirk-thomas: The user still needs to call a function from it to create a logger instance. So what gain would we get from an implicit import during init?

dirk-thomas:
I would expect a node to log its messages on a logger which uses the node name as the logger name.

Optionally with the package name in the front, but even that could be argued against with the same reasoning: if a user forks pkg A to pkg B it couldn't simply run the node from pkg B because all the logger names would be different again.

wjwwood:
It would be cool if the package name (possibly other things) would be part of the logger "instance" but not part of the logger name, i.e. something like a set of key-value pairs. These could be used in our logging output format, e.g. `[{package_name}/{logger_name}]: {msg}`. But that's not necessary. Users could include the package name in their own messages for now with https://github.com/ros2/ament_cmake_ros/pull/2.

If we assume that node names will be unique at some point, the name should be sufficient to ensure no cross talk in logging and therefore the package name isn't necessary.

rclpy/executors/examples_rclpy_executors/callback_group.py
dhood: that the diff for this PR would not include this line.

Without the implicit import, even users that just want to use the root logger have to import `rclpy.logging` to call `rclpy.logging.loginfo()` (unlike in rospy). It seems convenient to always have it imported but I am asking because it might not be good from a design perspective

dirk-thomas: My personal opinion is "import what you use" but that is certainly different from what is done in `rospy`. I will leave it up to others to give feedback on this.

wjwwood: I don't care one way or the other. I wouldn't be surprised/shocked to see that `import rclpy` allows me to do `rclpy.logging.*`, which is how `import os; os.path.join(...)` works. On the other hand, it would be fine to have to explicitly import logging too.

On balance, I'd slightly prefer not needing this import so the user's code is slightly shorter and I figure it will be a common enough thing to save a non-trivial amount of typing on user's part.

PR #187, opened 2017-10-30
Control shared/static linking via BUILD_SHARED_LIBS and CMAKE_TOOLCHAIN_FILE

lmayencourt:
This PR will allow the static linking of the example applications. 

minimal_composition is not added as it have more dependencies with
others ROS2 packages.

connects to ros2/ros2#306

### DIFF
--- a/rclcpp/minimal_client/CMakeLists.txt
+++ b/rclcpp/minimal_client/CMakeLists.txt
@@ -10,11 +10,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-if((NOT BUILD_SHARED_LIBS) AND CMAKE_TOOLCHAIN_FILE)
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static -pthread")
-endif()
-
-find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_ros REQUIRED)
 find_package(example_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 
--- a/rclcpp/minimal_client/CMakeLists.txt
+++ b/rclcpp/minimal_client/CMakeLists.txt
@@ -10,6 +10,10 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+if((NOT BUILD_SHARED_LIBS) AND CMAKE_TOOLCHAIN_FILE)
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static -pthread")
+endif()
+
 find_package(ament_cmake REQUIRED)
 find_package(example_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
--- a/rclcpp/minimal_client/package.xml
+++ b/rclcpp/minimal_client/package.xml
@@ -7,7 +7,7 @@
   <maintainer email="morgan@osrfoundation.org">Morgan Quigley</maintainer>
   <license>Apache License 2.0</license>
 
-  <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
 
   <build_depend>rclcpp</build_depend>
   <build_depend>example_interfaces</build_depend>
--- a/rclcpp/minimal_publisher/CMakeLists.txt
+++ b/rclcpp/minimal_publisher/CMakeLists.txt
@@ -10,6 +10,10 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+if((NOT BUILD_SHARED_LIBS) AND CMAKE_TOOLCHAIN_FILE)
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static -pthread")
+endif()
+
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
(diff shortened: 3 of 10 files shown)


mikaelarguedas:
Tip: @lmayencourt This is related to https://github.com/ros2/ros2/issues/306, so I'm adding the keyword "connects to ros2/ros2#306" in the first comment so that the issues are cross-referenced on github and connected on the [waffle board](waffle.io/ros2/ros2)

lmayencourt:
I was looking for this ! thanks

mikaelarguedas:
@lmayencourt could you give more context about the approach taken here please ?
I guess `ament_cmake_ros` cannot be used because what we want to do here is to link the executables statically. But I'm not sure why we have to provide a CMAKE_TOOLCHAIN_FILE to link statically

lmayencourt:
the idea here is to use the `BUILD_SHARED_LIBS` flag to add the appropriate flags to the linker (`-static -pthread`). 

>But I'm not sure why we have to provide a CMAKE_TOOLCHAIN_FILE to link statically


The `CMAKE_TOOLCHAIN_FILE` is only here to prevent a static compilation for x86. I tried and the behaviour is strange... `minimal_subscriber` and `minimal_client` are working fine, but `minimal_publisher` and `minimal_service` crashed... 
Also the use of a static libfastrtps don't allow you to run more than one example:
```
terminate called after throwing an instance of 'std::system_error'
  what():  bind: Address already in use
Abort (core dumped)
```
More time should be invested in the x86 side...

>I guess ament_cmake_ros cannot be used

I didn't think about `ament_cmake_ros` for this, I will have a look today !

lmayencourt:
Ok, so it is also possible to put the logic to add the linker flags on `ament_cmake_ros`. It is probably a better approach ! 
If you agree I can push a new patch that makes the examples depend on `ament_cmake_ros` instead of `ament_cmake` and rename this PR to " Control shared/static linking via ament_cmake_ros"

mikaelarguedas:
> If you agree I can push a new patch that makes the examples depend on ament_cmake_ros instead of ament_cmake and rename this PR to " Control shared/static linking via ament_cmake_ros"

:+1: sounds good to me

> The CMAKE_TOOLCHAIN_FILE is only here to prevent a static compilation for x86.

I'm not sure that relying on an unrelated variable is the best approach here. If I request a compilation of my workspace with static linking, I would expect it to link my libraries and executables statically regardless of platform / architecture. So I'd suggest investigating the behavior on x86 before considering merging this.

lmayencourt:
changing to `ament_cmake_ros` here should not be an issue. If we don't add extra linker flags in `ament_cmake_ros` for x86 we will obtain something like this:
```
#file install/lib/examples_rclcpp_minimal_publisher/publisher_lambda 
install/lib/examples_rclcpp_minimal_publisher/publisher_lambda: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=d43728c683e6dbe90104d87a2270aa4c5f059ae6, not stripped
```
```
#ldd install/lib/examples_rclcpp_minimal_publisher/publisher_lambda
	linux-vdso.so.1 =>  (0x00007ffc86dfe000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f7203e87000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f7203b05000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f72037fc000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f72035e6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f720321c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f72040a4000)
```
All the ROS2 libraries are linked statically as expected and the behaviour of the application is normal (I can run multiple instance of it without the issue describe above).

The use of `CMAKE_TOOLCHAIN_FILE` is maybe not the best option, but I think it's a good things to be able to specify linker flags according to the target architecture. Apparently the list of required flags is different between x86 and aarch64...
But I agree that the behavior on x86 for a fully static application need more investigation.

dirk-thomas:
@lmayencourt Since neither of these packages actually builds libraries I don't see what the change of the dependency would do. It shouldn't have any effect. So can you please clarify what the purpose of the patch is?

---

Just for the record: static linking of dependencies is being addressed in ros2/ros2#468. But it doesn't touch this repo currently.

lmayencourt:
The modification here was to allow to build a fully static binary according to the `BUILD_SHARED_LIB` flag.
With the modification of https://github.com/ros2/ros2/issues/468 you build ROS2 statically, but your are still using dynamic system libraries. You must add `-static -pthread` to the linker flags to eliminate this last dependencies. 
The last idea was to include this functionality in `ament_cmake_ros`

dirk-thomas:
Currently `ament_cmake_ros` doesn't have any such feature. So am I right that the current patch doesn't do anything? If that is the case I would suggest to open this when it is ready - or we can mark it with "requires changes" until the patch has actually an effect. Whichever option you prefer.

lmayencourt:
I'm alright with both option. 
I will have a look at `ament_cmake_ros` to see if I can add the required features.

dirk-thomas:
I will close the PR then for now. Once it actually results in a change it can be reopened.

esteve:
@dirk-thomas @lmayencourt adding those flags to `ament_cmake_ros` would certainly make the iOS build fail, since it only needs for the ROS2 libraries to be built statically, but not the C library.

Moreover, I don't see the rationale for statically linking the C library, it would just make executables much larger and with no added benefit. If the goal is to have completely self-contained executables, that wouldn't work since the C library is highly dependent on the version of the target kernel (syscalls need to match, for example). In any case, if a developer wants to build a fully statically ROS2 application, wouldn't it make more sense to do that in his/her own package instead of adding it to ROS2 and forcing that behavior to other packages?

lmayencourt:
@esteve you are right, since we already have all the ROS2 stack statically linked, we can expect that the target have the required dependencies to run the binary (pthread, ld, c and math library, ...). 
And it make probably more sense to add this specific flags directly in the package. 
@dirk-thomas  do you think It would be useful to have an example with the flags set ?

dirk-thomas:
Maybe an example helps in understanding the case and making a decision. Currently I don't have an answer.

PR #188, opened 2017-11-16
Switch to using rate

dhood:
Otherwise no example is covering it anymore

connects to ros2/demos#192

### DIFF
--- a/rclcpp/minimal_publisher/not_composable.cpp
+++ b/rclcpp/minimal_publisher/not_composable.cpp
@@ -30,12 +30,14 @@ int main(int argc, char * argv[])
   auto publisher = node->create_publisher<std_msgs::msg::String>("topic");
   auto message = std::make_shared<std_msgs::msg::String>();
   auto publish_count = 0;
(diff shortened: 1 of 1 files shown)


rclcpp/minimal_publisher/not_composable.cpp
```diff
@@ -30,12 +30,14 @@ int main(int argc, char * argv[])
   auto publisher = node->create_publisher<std_msgs::msg::String>("topic");
   auto message = std::make_shared<std_msgs::msg::String>();
   auto publish_count = 0;
+  rclcpp::WallRate loop_rate(2s);
```
wjwwood: This is 0.5Hz, but before it was 2Hz, right?

dhood: oops: e21d2d6f2414388235dd21ef7e4bbd907fc2bd3e puts it back to 2Hz

dhood:
:ship: :+1: :tada: :tada:

PR #190, opened 2017-12-05
Use logging

mikaelarguedas:
I also changed the message printing from `Publishing: [%s]` to `Publishing: '%s'` as 
`[INFO] [publisher_node]: Publisher: [Hello, world! 0]` was getting a bit full of `[` and harder to scope the user printed message.
In Python it prints  `Publishing: "%s"` but I can change it if we want the messages to be identical in both languages

### DIFF
--- a/rclcpp/minimal_service/main.cpp
+++ b/rclcpp/minimal_service/main.cpp
@@ -25,7 +25,9 @@ void handle_service(
   const std::shared_ptr<AddTwoInts::Response> response)
 {
   (void)request_header;
-  printf("request: %" PRId64 " + %" PRId64 "\n", request->a, request->b);
+  RCLCPP_INFO(
+    rclcpp::get_logger("minimal_service"),
+    "request: %" PRId64 " + %" PRId64, request->a, request->b)
   response->sum = request->a + request->b;
 }
 
--- a/rclcpp/minimal_service/main.cpp
+++ b/rclcpp/minimal_service/main.cpp
@@ -18,6 +18,7 @@
 #include "rclcpp/rclcpp.hpp"
 
 using AddTwoInts = example_interfaces::srv::AddTwoInts;
+rclcpp::Node::SharedPtr g_node = nullptr;
 
 void handle_service(
   const std::shared_ptr<rmw_request_id_t> request_header,
@@ -26,7 +27,7 @@ void handle_service(
 {
   (void)request_header;
   RCLCPP_INFO(
-    rclcpp::get_logger("minimal_service"),
+    g_node->get_logger(),
     "request: %" PRId64 " + %" PRId64, request->a, request->b)
   response->sum = request->a + request->b;
 }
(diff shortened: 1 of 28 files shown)


mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3721/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/879/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3058/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3817/)

Though CI wont show much as examples are not tested purposefully

rclcpp/minimal_service/main.cpp
```diff
@@ -25,7 +25,9 @@ void handle_service(
   const std::shared_ptr<AddTwoInts::Response> response)
 {
   (void)request_header;
-  printf("request: %" PRId64 " + %" PRId64 "\n", request->a, request->b);
+  RCLCPP_INFO(
+    rclcpp::get_logger("minimal_service"),
```
dhood: for free-function callbacks we've been capturing the logger in a callback and passing it to the handler: https://github.com/ros2/demos/pull/191/commits/52dbdac6dd29bfaab4cf938264bf43c8ffba46a8

mikaelarguedas: I thought this was unnecessary complexity here for just passing a string around so I didnt do it on purpose.

dhood: Which string so you mean? If it's the logger name then the purpose is to have it reflect remapping on the node name (which isn't possible today but they'd be set up for when it's possible)

mikaelarguedas: I meant the logger. I think that as it stands it doesnt deserve doubling the size of the main to propagate it. I integrated @wjwwood suggestion (#191) instead.

> If it's the logger name then the purpose is to have it reflect remapping on the node name

Yeah indeed that's important to leverage when this becomes available. I was just discussing this with wjwwood and we agreed that hopefully by then we will provide an easy way for user to pass additional argument to callbacks as logging from callback is something very common.

mikaelarguedas:
ready for review

dhood:
@mikaelarguedas to avoid conflicts with https://github.com/ros2/examples/pull/192 I removed 71d215e9d413315115e0ec447005c08ffd18c107

rclcpp/minimal_service/main.cpp
```diff
@@ -18,23 +18,26 @@
...
 
 void handle_service(
   const std::shared_ptr<rmw_request_id_t> request_header,
   const std::shared_ptr<AddTwoInts::Request> request,
   const std::shared_ptr<AddTwoInts::Response> response)
 {
   (void)request_header;
-  printf("request: %" PRId64 " + %" PRId64 "\n", request->a, request->b);
+  RCLCPP_INFO(
+    g_node->get_logger(),
```
dirk-thomas: Segfaults since the global variable is never set.

mikaelarguedas: Actually this example works fine. I think you're referring to the not composable subscriber. See #195 for a fix

PR #191, opened 2017-12-05
 possible simple solution to free function logging

wjwwood:
You don't have to use this, but just an option if you want it.


mikaelarguedas:
Yeah that's a better alternative to keep the main minimal. I'll do that for the not_composable examples as well :+1:

PR #195, opened 2017-12-06
use global node not local

mikaelarguedas:
forgot to remove the auto in #190


mikaelarguedas:
added a second commit to fix segfaults on exit

PR #197, opened 2017-12-17
Use std::chrono_literals only

youtalk:
All examples in `rclcpp` directory use `std::chrono_literals` except in `publisher_node.cpp`.
This PR changes to use `500ms` instead of `std::chrono::milliseconds(500)` using `std::chrono_literals` such as the other examples.

### DIFF
--- a/rclcpp/minimal_composition/src/publisher_node.cpp
+++ b/rclcpp/minimal_composition/src/publisher_node.cpp
@@ -18,12 +18,14 @@
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/string.hpp"
 
+using namespace std::chrono_literals;
+
 PublisherNode::PublisherNode()
 : Node("publisher_node"), count_(0)
 {
   publisher_ = create_publisher<std_msgs::msg::String>("topic");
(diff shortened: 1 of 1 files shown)


rclcpp/minimal_composition/src/publisher_node.cpp
```diff
@@ -12,18 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <chrono>
-
```
dirk-thomas: Since the code is using `std::chrono_literals` it needs to keep this include.

youtalk: @dirk-thomas Thank you for reviewing. I've reverted it.

Because other cpp files don't also include it, I remove it.

dirk-thomas:
Thank you for the patch.

> Because other cpp files don't also include it, I remove it.

Everything a source file uses should be explicitly included since the source file can't assume that other header don't change in the future and might remove the transitive dependency.

youtalk:
That sounds good. I'll make another PR to make it.

PR #203, opened 2018-02-02
Update examples for new client api

sloretz:
connects to ros2/rclpy#170

`client_async_callback` requires bug fix in ros2/rclpy#176


mjcarroll:
Is there somewhere that explains how to run these?

mikaelarguedas:
I don't think so:
`ros2 run examples_rclpy_<TAB><TAB>` should give you the lost of rclpy example packages and then allow you to autocomplete the executable names

Issue #209, opened 2018-06-21
subscriber_not_composable always throws an error after Ctrl-C


clalancette:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Bionic arm64
- Installation type:
  - Fat archive from 2018-06-21, downloaded from here: https://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/475/artifact/ws/ros2-package-linux-aarch64.tar.bz2
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
$ ros2 run examples_rclcpp_minimal_subscriber subscriber_not_composable
```

Wait a few seconds, then press Ctrl-C gives:

```
^Csignal_handler(2)
[ERROR] [rclcpp]: Error in destruction of rcl subscription handle: the Node Handle was destructed too early. You will leak memory
[ERROR] [rclcpp]: Error in destruction of rcl subscription handle: the Node Handle was destructed too early. You will leak memory
```

#### Expected behavior
Program quits without error.

#### Actual behavior
Program quits with: `[ERROR] [rclcpp]: Error in destruction of rcl subscription handle: the Node Handle was destructed too early. You will leak memory`

#### Additional information
This bug may be in the underlying `rclcpp` or `rcl` layers, but since I wasn't sure, I started here.

dhood:
I have also seen this with a different executable using Connext. It's not every time so might be a race condition:

```

$ ros2 run  intra_process_demo image_pipeline_all_in_one
RTI Data Distribution Service EVAL License issued to OSRF dhood@osrfoundation.org For non-production use only.
Expires on 16-Jul-2018 See www.rti.com for more information.
RTI Data Distribution Service EVAL License issued to OSRF dhood@osrfoundation.org For non-production use only.
Expires on 16-Jul-2018 See www.rti.com for more information.
RTI Data Distribution Service EVAL License issued to OSRF dhood@osrfoundation.org For non-production use only.
Expires on 16-Jul-2018 See www.rti.com for more information.
^Csignal_handler(2)
PRESInterParticipant_removeRemoteParticipant:[Local Participant: ac1701a8 681b 1] [Remote Participant: 1c1 10f01a8 71680000 | Remote Endpoint: 0] Could not remove remote endpoint
[...]

PR #210, opened 2018-06-22
Change the not_composable example to destroy subscription first.

clalancette:
We are seeing the following message when hitting Ctrl-C
in the subscriber not_composable node:

[ERROR] [rclcpp]: Error in destruction of rcl subscription handle: the Node Handle was destructed too early. You will leak memory

The problem is that setting g_node to nullptr causes the
node destructor to be run before the subscriber is destructed.
Due to another bug, we can't just remove that assignment, so
instead we make it all locally scoped in main for now, which
works around the problem.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This worksaround https://github.com/ros2/examples/issues/209

### DIFF
--- a/rclcpp/minimal_subscriber/not_composable.cpp
+++ b/rclcpp/minimal_subscriber/not_composable.cpp
@@ -15,8 +15,6 @@
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/string.hpp"
 
-rclcpp::Node::SharedPtr g_node = nullptr;
-
 /* We do not recommend this style anymore, because composition of multiple
  * nodes in the same executable is not possible. Please see one of the subclass
  * examples for the "new" recommended styles. This example is only included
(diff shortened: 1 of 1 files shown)


clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4795/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1656/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3977/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4849/)

rclcpp/minimal_subscriber/not_composable.cpp
```diff
@@ -15,26 +15,26 @@
...
-
 /* We do not recommend this style anymore, because composition of multiple
  * nodes in the same executable is not possible. Please see one of the subclass
  * examples for the "new" recommended styles. This example is only included
  * for completeness because it is similar to "classic" standalone ROS nodes. */
 
 void topic_callback(const std_msgs::msg::String::SharedPtr msg)
 {
-  RCLCPP_INFO(g_node->get_logger(), "I heard: '%s'", msg->data.c_str())
+  RCLCPP_INFO(rclcpp::get_logger("minimal_subscriber"), "I heard: '%s'", msg->data.c_str())
```
wjwwood: This is not the same thing. This function actually does need the node instance to maintain the existing behavior (and the existing behavior is the desired behavior in my opinion).

clalancette: All right, in that case, I've changed it to manually destroy the subscription first, which achieves the same effect.  I've also updated the comment.  Please take another look when you get a chance.

clalancette:
CI for the new code:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4806/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1667/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3991/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4860/)

PR #215, opened 2018-10-08
Set library path environment hook.

nuclearsandwich:
Set library path environment hook for this package. This package is one of a handful which installs libraries without setting an environment hook to add them to the platform library path.

Connects to ros2/ros_workspace#10

### DIFF
--- a/rclcpp/minimal_composition/CMakeLists.txt
+++ b/rclcpp/minimal_composition/CMakeLists.txt
@@ -24,6 +24,14 @@ target_compile_definitions(composition_nodes
   PRIVATE "MINIMAL_COMPOSITION_DLL")
 ament_target_dependencies(composition_nodes rclcpp std_msgs class_loader)
 
+# This package installs libraries without exporting them.
+# Export the library path to ensure that the installed libraries are available.
+if(NOT WIN32)
+  ament_environment_hooks(
+    "${ament_cmake_package_templates_ENVIRONMENT_HOOK_LIBRARY_PATH}"
+    )
+endif()
+
 add_executable(composition_publisher src/standalone_publisher.cpp)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Is the environment variable required for this package (when it only loads these components through pluginlib)?

nuclearsandwich:
> Is the environment variable required for this package (when it only loads these components through pluginlib)?

Yeah it still needs to be on the load path. It's pretty easy to verify this in an isolated install space:
```
$ . install/setup.bash 
$ ros2 run examples_rclcpp_minimal_composition composition_composed
[INFO] [publisher_node]: Publisher: 'Hello, world! 0'
[INFO] [subscriber_node]: Subscriber: 'Hello, world! 0'
[INFO] [publisher_node]: Publisher: 'Hello, world! 1'
[INFO] [subscriber_node]: Subscriber: 'Hello, world! 1'
^Csignal_handler(2)
$ ^C
$ LD_LIBRARY_PATH=$(echo $LD_LIBRARY_PATH | tr ':' '\n' | grep -v examples_rclcpp_minimal_composition | tr '\n' ':')
$ ros2 run examples_rclcpp_minimal_composition composition_composed
/home/steven/osrf/environment_hooks_sweep_ws/install/examples_rclcpp_minimal_composition/lib/examples_rclcpp_minimal_composition/composition_composed: error while loading shared libraries: libcomposition_nodes.so: cannot open shared object file: No such file or directory
```

nuclearsandwich:
A similar change was made to the composition demo in https://github.com/ros2/demos/pull/183

nuclearsandwich:
CI for this change was run successfully and linked in https://github.com/ros2/ros_workspace/pull/10#issuecomment-428005194

PR #222, opened 2019-01-09
rclpy action examples

jacobperron:
Replaces #216.

PR #226, opened 2019-02-02
Make examples install entry point scripts into a package

sloretz:
This makes the examples install their scripts into a package with their name. Currently they install their modules on their own, which is a problem since some of the modules have generic names.

See https://github.com/ros2/examples/pull/222#issuecomment-459918099 for a bug this caused. `import client` would import the minimal client example. With this PR that import becomes `import examples_rclpy_minimal_client.client`.


sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6111/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2586/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5040/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5981/)

PR #227, opened 2019-02-12
update to use separated action types

dirk-thomas:
Connected to ros2/rosidl#334.

### DIFF
--- a/rclcpp/minimal_action_server/not_composable.cpp
+++ b/rclcpp/minimal_action_server/not_composable.cpp
@@ -49,8 +49,8 @@ void execute(
   RCLCPP_INFO(rclcpp::get_logger("server"), "Executing goal");
   rclcpp::Rate loop_rate(1);
   const auto goal = goal_handle->get_goal();
-  auto feedback = std::make_shared<Fibonacci::Impl::FeedbackMessage>();
-  auto& sequence = feedback->feedback.sequence;
+  auto feedback = std::make_shared<Fibonacci::Feedback>();
+  auto& sequence = feedback->sequence;
   sequence.push_back(0);
   sequence.push_back(1);
   auto result = std::make_shared<Fibonacci::Result>();
--- a/rclcpp/minimal_action_server/not_composable.cpp
+++ b/rclcpp/minimal_action_server/not_composable.cpp
@@ -53,15 +53,15 @@ void execute(
   auto& sequence = feedback->feedback.sequence;
   sequence.push_back(0);
   sequence.push_back(1);
-  auto result_response = std::make_shared<Fibonacci::Impl::GetResultService::Response>();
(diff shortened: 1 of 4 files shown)


rclcpp/minimal_action_server/not_composable.cpp
```diff
@@ -49,18 +49,18 @@ void execute(
   RCLCPP_INFO(rclcpp::get_logger("server"), "Executing goal");
   rclcpp::Rate loop_rate(1);
   const auto goal = goal_handle->get_goal();
-  auto feedback = std::make_shared<Fibonacci::Feedback>();
-  auto& sequence = feedback->sequence;
+  auto feedback = std::make_shared<Fibonacci::Impl::FeedbackMessage>();
```
hidmic: @dirk-thomas this seems off, the user shouldn't have to fiddle with implementation details.

rclcpp/minimal_action_server/not_composable.cpp
```diff
@@ -49,18 +49,18 @@ void execute(
...
+  auto& sequence = feedback->feedback.sequence;
   sequence.push_back(0);
   sequence.push_back(1);
-  auto result_response = std::make_shared<Fibonacci::Result>();
+  auto result_response = std::make_shared<Fibonacci::Impl::GetResultService::Response>();
```
hidmic: @dirk-thomas same concern as above.

dirk-thomas: `Maybe you can propose how to achieve this in the light of calling `set_canceled` below?

sloretz: `Fibonacci::Result` was done with a using directive on the generated `Fibonacci` class. 
https://github.com/ros2/rosidl/blob/7c5c7d3af98fb8f5fd8ae9112ee278d21f899df1/rosidl_generator_cpp/resource/action__struct.hpp.em#L45

The equivalent would be

```C++
using Result = Impl::GetResultService::Response;
```

dirk-thomas: Isn't that against the goal to not expose those wrapping types to the user? I thought that is what @hidmic's comment is about.

dirk-thomas: Resolved in 3c2b2e804c1395cdb289365291b30b2050ddb610 after changing the API signature in `rclcpp_action`.

rclcpp/minimal_action_server/not_composable.cpp
dirk-thomas: Resolved in 3c2b2e804c1395cdb289365291b30b2050ddb610 after changing the API signature in `rclcpp_action`.

PR #230, opened 2019-03-12
Add rclcpp action examples with member functions

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6398/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2785/) (unrelated failure)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5262/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6235/)

Connects to ros2/rclcpp#635

### DIFF
--- a/rclcpp/minimal_action_client/CMakeLists.txt
+++ b/rclcpp/minimal_action_client/CMakeLists.txt
@@ -15,6 +15,12 @@ find_package(example_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_action REQUIRED)
 
+add_executable(action_client_member_functions member_functions.cpp)
+ament_target_dependencies(action_client_member_functions
+  "rclcpp"
+  "rclcpp_action"
+  "example_interfaces")
+
 add_executable(action_client_not_composable not_composable.cpp)
 ament_target_dependencies(action_client_not_composable
   "rclcpp"
@@ -34,6 +40,7 @@ ament_target_dependencies(action_client_not_composable_with_feedback
   "example_interfaces")
 
 install(TARGETS
+  action_client_member_functions
   action_client_not_composable
   action_client_not_composable_with_cancel
   action_client_not_composable_with_feedback
--- a/rclcpp/minimal_action_client/member_functions.cpp
+++ b/rclcpp/minimal_action_client/member_functions.cpp
@@ -1,0 +1,143 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <inttypes.h>
+#include <memory>
+#include <string>
+#include <iostream>
+#include "example_interfaces/action/fibonacci.hpp"
+#include "rclcpp/rclcpp.hpp"
+// TODO(jacobperron): Remove this once it is included as part of 'rclcpp.hpp'
+#include "rclcpp_action/rclcpp_action.hpp"
+
+class MinimalActionClient : public rclcpp::Node
+{
+public:
+  using Fibonacci = example_interfaces::action::Fibonacci;
+  using GoalHandleFibonacci = rclcpp_action::ClientGoalHandle<Fibonacci>;
+
+  MinimalActionClient(const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions())
+  : Node("minimal_action_client", node_options), goal_done_(false)
+  {
+    this->client_ptr_ = rclcpp_action::create_client<Fibonacci>(
+      this->get_node_base_interface(),
+      this->get_node_graph_interface(),
+      this->get_node_logging_interface(),
+      this->get_node_waitables_interface(),
+      "fibonacci");
+
+    this->timer_ = this->create_wall_timer(
+      std::chrono::milliseconds(500),
+      std::bind(&MinimalActionClient::send_goal, this));
+  }
+
+  bool is_goal_done() const
+  {
+    return this->goal_done_;
+  }
+
+  void send_goal()
+  {
+    using namespace std::placeholders;
+
+    this->timer_->cancel();
+
+    this->goal_done_ = false;
+
+    if (!this->client_ptr_) {
+      RCLCPP_ERROR(this->get_logger(), "Action client not initialized");
+    }
+
+    if (!this->client_ptr_->wait_for_action_server(std::chrono::seconds(10))) {
+      RCLCPP_ERROR(this->get_logger(), "Action server not available after waiting");
+      this->goal_done_ = true;
+      return;
+    }
+
+    auto goal_msg = Fibonacci::Goal();
+    goal_msg.order = 10;
+
+    RCLCPP_INFO(this->get_logger(), "Sending goal");
+
+    auto send_goal_options = rclcpp_action::Client<Fibonacci>::SendGoalOptions();
+    send_goal_options.goal_response_callback =
+      std::bind(&MinimalActionClient::goal_response_callback, this, _1);
+    send_goal_options.feedback_callback =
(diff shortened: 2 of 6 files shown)


rclcpp/minimal_action_client/member_function.cpp
```diff
@@ -0,0 +1,123 @@
...
+      feedback->sequence.back());
+  }
+};  // class MinimalActionClient
+
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  auto action_client = std::make_shared<MinimalActionClient>("minimal_action_client");
+
+  action_client->spin_until_goal_complete(10);
```
sloretz: This works, but it would be difficult to compose because it needs to call the executor to block. It would be easier if there was to run a function when a future completes.

jacobperron: I guess we should add callbacks as described in https://github.com/ros2/rclcpp/issues/635 to make this example better.

jacobperron: I've refactored the composable example to use callbacks in 82c7acb. This depends on ros2/rclcpp#701.

rclcpp/minimal_action_client/member_function.cpp
```diff
@@ -0,0 +1,137 @@
...
+public:
+  using Fibonacci = example_interfaces::action::Fibonacci;
+  using GoalHandleFibonacci = rclcpp_action::ClientGoalHandle<Fibonacci>;
+
+  MinimalActionClient(const std::string node_name)
+    : Node(node_name), goal_done_(false)
+  {
+  }
+
+  void init(const std::string action_name)
```
Karsten1987: I don't think this will work properly with composition. The component manager can't call functions on the class but really only initializes them (AFAICT).

rclcpp/minimal_action_client/member_function.cpp
```diff
@@ -0,0 +1,137 @@
...
+  {
+    this->client_ptr_ =
+      rclcpp_action::create_client<Fibonacci>(this->shared_from_this(), action_name);
```
Karsten1987: Instead of `shared_from_this()`, you can use the node interfaces for creating a server/client:
https://github.com/ros2/rclcpp/blob/master/rclcpp_action/include/rclcpp_action/create_server.hpp#L56-L68

With this, you can put the logic into the constructor. An example can be found here:
https://github.com/Karsten1987/roscon2018/blob/composition_components/confbot_driver/src/nodes/confbot_driver.cpp#L41-L52

rclcpp/minimal_action_client/member_function.cpp
```diff
@@ -0,0 +1,137 @@
...
+// TODO(jacobperron): Remove this once it is included as part of 'rclcpp.hpp'
+#include "rclcpp_action/rclcpp_action.hpp"
+
+class MinimalActionClient : public rclcpp::Node
+{
+public:
+  using Fibonacci = example_interfaces::action::Fibonacci;
+  using GoalHandleFibonacci = rclcpp_action::ClientGoalHandle<Fibonacci>;
+
+  MinimalActionClient(const std::string node_name)
```
Karsten1987: composition requires a constructor with `rclcpp::NodeOptions`.

rclcpp/minimal_action_client/member_function.cpp
jacobperron: Great, thanks for https://github.com/ros2/rclcpp/pull/700 :)

rclcpp/minimal_action_client/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ find_package(example_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_action REQUIRED)
 
+add_executable(action_client_member_function member_function.cpp)
+ament_target_dependencies(action_client_member_function
+  "rclcpp"
+  "rclcpp_action"
+  "example_interfaces")
+
 add_executable(action_client_not_composable not_composable.cpp)
```
mjcarroll: I think this also a good opportunity to remove the non_composable variants.

jacobperron: My understanding is that this repo contains minimal examples for all variants, and those that are the recommended way to do things should be in [demos](https://github.com/ros2/demos) or [tutorials](https://github.com/ros2/tutorials) ([relevant ROS Answer](https://answers.ros.org/question/318089/difference-of-ros2-demos-and-ros2-examples/)). Although, it does seem a little redundant IMO.

jacobperron:
@Karsten1987 I've updated the examples to be more in line with the new composition model. Please take another look (thanks!).

rclcpp/minimal_action_client/CMakeLists.txt
mjcarroll: Fair enough.  I think that this may warrant a discussion at the next ROS2 meeting, because I was under the impression that the non-composable stuff can be dropped.

rclcpp/minimal_action_client/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ find_package(example_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_action REQUIRED)
```
mjcarroll: `find_package(rclcpp_components REQUIRED)`

rclcpp/minimal_action_client/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ find_package(example_interfaces REQUIRED)
...
 
+add_executable(action_client_member_functions member_functions.cpp)
+ament_target_dependencies(action_client_member_functions
+  "rclcpp"
+  "rclcpp_action"
+  "example_interfaces")
```
mjcarroll: "rclcpp_components"

mjcarroll: rclcpp_components_register_nodes(action_client_member_functions "MinimalActionClient")

rclcpp/minimal_action_client/member_functions.cpp
```diff
@@ -0,0 +1,138 @@
...
+        return;
+    }
+
+    RCLCPP_INFO(this->get_logger(), "Result received");
+    for (auto number : result.result->sequence) {
+      RCLCPP_INFO(this->get_logger(), "%" PRId64, number);
+    }
+  }
+};  // class MinimalActionClient
+
```
mjcarroll:
```
#include "rclcpp_components/register_node_macro.hpp"
RCLCPP_COMPONENTS_REGISTER_NODE(MinimalActionClient)
```

jacobperron:
@mjcarroll Thanks for the review!

In the spirit of keeping the examples in this package minimal, I'd prefer not to make these actual components and bring in a dependency to `rclcpp_component`. They are really examples of action clients/servers that use member functions, though I've tried to set them up so they could easily be ported to be components. I think it would be better to add a components version to `demos` or even `examples/composition`.

wjwwood:
I'll just add that, our long term goal has been to use composition everywhere, and make things that have their own main as the "exceptional cases". However, I can see an argument for not doing this just here, but instead leaving this pr as-is and then updating all the examples at once.

This is probably a topic we should discuss at some point before the release...

rclcpp/minimal_action_client/CMakeLists.txt
jacobperron: I've added it to the agenda.

Karsten1987:
@jacobperron This is still not fully composable. Each node has to be its own library and the node has to be registered with the class loader macro
```
#include "rclcpp_components/register_node_macro.hpp"
RCLCPP_COMPONENTS_REGISTER_NODE(MinimalActionClient)
```
as well as added to the ament index within the cmake context:
```
rclcpp_components_register_nodes(action_client_member_functions "MinimalActionClient")
```

Also, in order to support manual composition - besides the command line tool via `ros2 component` - I think it makes sense to separate the node library in a `.hpp` and `.cpp`.

jacobperron:
Based on our offline discussion, I'll leave these examples as not fully composable and create a follow-up PR to add composable examples to the [demos](https://github.com/ros2/demos) repo.

I still need to update this PR based on the changes in ros2/rclcpp#701.

jacobperron:
Ready for another round of review.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6776/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3077/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5572/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6604/)

PR #232, opened 2019-03-26
Update minimal_composition for rclcpp_components

mjcarroll:
Update to take advantage of the new rclcpp_components.

Connects to ros2/rclcpp#665


rclcpp/minimal_composition/include/minimal_composition/publisher_node.hpp
```diff
@@ -15,14 +15,16 @@
 #ifndef MINIMAL_COMPOSITION__PUBLISHER_NODE_HPP_
 #define MINIMAL_COMPOSITION__PUBLISHER_NODE_HPP_
 
+#include <string>
```
sloretz: nit, it doesn't look like this header is needed.

PR #239, opened 2019-05-05
avoid deprecated publish signature

wjwwood:
Connects to ros2/rclcpp#709


rclcpp/minimal_publisher/not_composable.cpp
```diff
@@ -29,13 +29,13 @@ int main(int argc, char * argv[])
   rclcpp::init(argc, argv);
   auto node = rclcpp::Node::make_shared("minimal_publisher");
   auto publisher = node->create_publisher<std_msgs::msg::String>("topic");
-  auto message = std::make_shared<std_msgs::msg::String>();
+  auto message = std_msgs::msg::String();
```
ivanpauno: I think that `std_msgs::msg::String message` is the usual style.

wjwwood: 👍 good call, updated

PR #240, opened 2019-05-07
Avoid deprecated API's by providing history settings

wjwwood:
I also ran `ament_uncrustify` and `ament_cpplint` over these, since it seems it's not run continually.

Connects to ros2/rclcpp#713


jacobperron:
It's probably worth adding linter tests in a follow-up.

PR #249, opened 2019-07-25
[dashing_backport] make member function const (#247)

Karsten1987:
Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/rclcpp/minimal_subscriber/member_function.cpp
+++ b/rclcpp/minimal_subscriber/member_function.cpp
@@ -29,7 +29,7 @@ public:
   }
 
 private:
-  void topic_callback(const std_msgs::msg::String::SharedPtr msg)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The definition is in a `.cpp` file only. So it is not part of the public API / ABI - no external package can rely on it.

jacobperron:
> The definition is in a .cpp file only. So it is not part of the public API / ABI - no external package can rely on it.

Missed that. Thanks. :upside_down_face:

jacobperron:
I expect CI will depend on https://github.com/ros2/rclcpp/pull/796

jacobperron:
I included this PR as part of the CI job for https://github.com/ros2/rclcpp/pull/796#issuecomment-516059028.

PR #250, opened 2019-08-01
[Backport] Make member function const

nuclearsandwich:
Backports #247 to Dashing, is a repeat of #249 targeting the `dashing` branch.

### DIFF
--- a/rclcpp/minimal_subscriber/member_function.cpp
+++ b/rclcpp/minimal_subscriber/member_function.cpp
@@ -29,7 +29,7 @@ public:
   }
 
 private:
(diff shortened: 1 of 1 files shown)


sloretz:
@nuclearsandwich DCO check is failing, but I'm not sure this is ABI compatible. While the function was private, since it's used from the constructor and the constructor is defined in the header the compiler could have inlined it. If so, code built against this will fail to find the method since `const` changes the mangled name.

nuclearsandwich:
@sloretz thanks for the raising the question. The ABI discussion was brought up in #249 and the determination was that because this function definition is in a cpp file it's not part of the public "ABI/API".

PR #253, opened 2019-10-21
Update client examples to expect future.result() to raise if a service call fails

sloretz:
This is the correct API since https://github.com/ros2/rclpy/pull/412 , which is new in Eloquent.

### DIFF
--- a/rclpy/services/minimal_client/examples_rclpy_minimal_client/client.py
+++ b/rclpy/services/minimal_client/examples_rclpy_minimal_client/client.py
@@ -31,12 +31,14 @@ def main(args=None):
     future = cli.call_async(req)
     rclpy.spin_until_future_complete(node, future)
 
-    if future.result() is not None:
+    try:
+        result = future.result()
+    except Exception as e:
+        node.get_logger().info('Service call failed %r' % (e,))
+    else:
         node.get_logger().info(
             'Result of add_two_ints: for %d + %d = %d' %
-            (req.a, req.b, future.result().sum))
-    else:
-        node.get_logger().info('Service call failed %r' % (future.exception(),))
+            (req.a, req.b, result.sum))
 
     node.destroy_node()
     rclpy.shutdown()
--- a/rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async.py
+++ b/rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async.py
@@ -34,12 +34,14 @@ def main(args=None):
     while rclpy.ok():
         rclpy.spin_once(node)
         if future.done():
-            if future.result() is not None:
+            try:
+                result = future.result()
+            except Exception as e:
+                node.get_logger().info('Service call failed %r' % (e,))
+            else:
                 node.get_logger().info(
                     'Result of add_two_ints: for %d + %d = %d' %
-                    (req.a, req.b, future.result().sum))
-            else:
-                node.get_logger().info('Service call failed %r' % (future.exception(),))
+                    (req.a, req.b, result.sum))
             break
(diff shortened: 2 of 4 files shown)


rclpy/services/minimal_client/examples_rclpy_minimal_client/client.py
```diff
@@ -31,12 +31,14 @@ def main(args=None):
     future = cli.call_async(req)
     rclpy.spin_until_future_complete(node, future)
 
-    if future.result() is not None:
+    try:
+        result = future.result()
+    except Exception as e:
+        node.get_logger().info('Service call failed %r' % (e,))
+    else:
```
jacobperron: nit: instead of a separate else-block, I would have preferred to move the successful log statement in try-block.

rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async.py
```diff
@@ -34,12 +34,14 @@ def main(args=None):
     while rclpy.ok():
         rclpy.spin_once(node)
         if future.done():
-            if future.result() is not None:
+            try:
+                result = future.result()
+            except Exception as e:
+                node.get_logger().info('Service call failed %r' % (e,))
+            else:
```
jacobperron: Same comment as above.

rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async_callback.py
```diff
@@ -37,13 +37,14 @@ def main(args=None):
             req.a = 41
             req.b = 1
             future = cli.call_async(req)
-            result = await future
-            if result is not None:
+            try:
+                result = await future
+            except Exception as e:
+                node.get_logger().info('Service call failed %r' % (e,))
+            else:
```
jacobperron: ditto

rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async_member_function.py
```diff
@@ -42,14 +42,15 @@ def main(args=None):
...
                 response = minimal_client.future.result()
+            except Exception as e:
                 minimal_client.get_logger().info(
-                    'Result of add_two_ints: for %d + %d = %d' %
-                    (minimal_client.req.a, minimal_client.req.b, response.sum))
+                    'Service call failed %r' % (e,))
             else:
                 minimal_client.get_logger().info(
-                    'Service call failed %r' % (minimal_client.future.exception(),))
+                    'Result of add_two_ints: for %d + %d = %d' %
```
jacobperron: ditto

rclpy/services/minimal_client/examples_rclpy_minimal_client/client.py
sloretz: I'd probably write it the way you mentioned in practice. If it seems reasonable to you, as an example I like that the try/except block is around the minimum amount of code. Maybe it would help someone reading the example understand what's expected to raise?

jacobperron: Sounds fine to me.

sloretz:
CI (testing just `examples_rclpy_minimal_client`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8500/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4385/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6917/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8410/)

PR #258, opened 2019-11-18
Fix client_cancel example

hidmic:
Precisely what the title says. Removes unused import of a nonexistent class.


hidmic:
There're no tests for this piece of code, not even linters.

PR #262, opened 2020-04-01
add examples for how to use the new rclcpp::WaitSet classes

wjwwood:
This is still WIP (not ready for review or comments).

Connects to https://github.com/ros2/rclcpp/pull/1047

### DIFF
--- a/rclcpp/wait_set/CMakeLists.txt
+++ b/rclcpp/wait_set/CMakeLists.txt
@@ -1,0 +1,24 @@
+cmake_minimum_required(VERSION 3.5)
+project(examples_rclcpp_wait_set)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(std_msgs REQUIRED)
+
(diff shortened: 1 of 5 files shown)


ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-05-21/14247/1

fred-apex-ai:
@wjwwood I just looked into the example code and it helped me quite  a bit. I was wondering if you plan to finish this PR, it has been on hold since spring. If you do, it would be great to add the arguably more common use case of a node that is spinning and how to use a waitset inside the node to wait for input from another node.
When it's spinning, it's easier to interact with the node, e.g. to query it `use_sim_time` is `true` etc.

wjwwood:
I want to, and now I might have the time, with help from others.

> If you do, it would be great to add the arguably more common use case of a node that is spinning and how to use a waitset inside the node to wait for input from another node.

That's a good idea. These examples were really "does it work" more than "this is what I expect will happen in the wild". But we should improve them to be more like the latter.

> When it's spinning, it's easier to interact with the node, e.g. to query it use_sim_time is true etc.

Yes, these examples were just testing the waitset part, but how it might be used in a larger scope.

nnmm:
Echoing Fred's request, I'm also wondering how you could transfer responsibility for a subscription from an executor to a waitset, especially when you don't control the executor (like in a component container).

It's not even clear to me how to instantiate a static wait set with a single subscription – I would expect this to work, but it doesn't (in Foxy):
```
// in the initializer list
wait_set_(std::array<rclcpp::StaticWaitSet<1, 0, 0, 0, 0, 0>::SubscriptionEntry, 1>{{sub_}})
```

wjwwood:
Closing in favor of https://github.com/ros2/examples/pull/315

PR #264, opened 2020-04-08
Restructure rclcpp folders

maryaB-osr:
This isn't very important, but the way the [rclpy packages](https://github.com/ros2/examples/tree/master/rclpy) are organized makes a nice landing place for the tutorials to point to based on the topic. 

So in the python pub/sub tutorial for example, we can link right to [rclpy/topics](https://github.com/ros2/examples/tree/master/rclpy/topics) and the reader will see exactly what they need to see and nothing extra. Whereas if I wanted to do the same for the c++ pub/sub tutorial, I have to link to just [rclcpp](https://github.com/ros2/examples/tree/master/rclcpp) and the reader is bombarded by all these other packages they don't need to see. That's my reasoning, but again, it's not super important.

Also I wasn't sure if `minimal_composition`, `minimal_timer` and `multithreaded_executor` could all be placed in one folder or what that folder would be called.

Signed-off-by: maryaB-osr <marya@openrobotics.org>


clalancette:
> So in the python pub/sub tutorial for example, we can link right to [rclpy/topics](https://github.com/ros2/examples/tree/master/rclpy/topics) and the reader will see exactly what they need to see and nothing extra. Whereas if I wanted to do the same for the c++ pub/sub tutorial, I have to link to just [rclcpp](https://github.com/ros2/examples/tree/master/rclcpp) and the reader is bombarded by all these other packages they don't need to see. That's my reasoning, but again, it's not super important.

Looking at this, I actually agree with your reasoning.  Putting in the subdirectories gives it a bit of extra clarity that I like.

> Also I wasn't sure if `minimal_composition`, `minimal_timer` and `multithreaded_executor` could all be placed in one folder or what that folder would be called.

No, they all deal with different aspects of what rclcpp can do.  Going with this new structure, I think the directory structure should be something like this:

```
- rclcpp
    - actions
        - minimal_action_client
        - minimal_action_server
    - composition
        - minimal_composition
    - executors
        - multithreaded_executor
    - services
        - minimal_client
        - minimal_service
    - timers
        - minimal_timer
    - topics
        - minimal_publisher
        - minimal_subscriber
```

Separately we should add a "timer" tutorial to the rclpy stuff, but that is follow-up work.

@maryaB-osr Does that make sense to you?

maryaB-osr:
@clalancette Yup that makes sense to me. "Time", "executors/spinners" and "composition" are all on the list for new tutorials :)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10279/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5815/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8360/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10175/)

PR #265, opened 2020-04-20
Added common linters

ahcorde:
Added common linters

Signed-off-by: ahcorde <ahcorde@gmail.com>


dirk-thomas:
Save question as on ros2/example_interfaces#10

ahcorde:
In this case as the description of this set of packages mentioned

```
This directory contains many examples of how to do common tasks with rclcpp.
The intent is that this material will be easy to copy-and-paste into your own projects.
```

Many user may copy-and-paste portions of this code in their project. I think that we need to assure in these packages that the quality of code is correct and provide them the right code from the very beginning.

Feel free to close it if you consider that these changes are not adding any value.

dirk-thomas:
That is exactly my question: why should a user call the linters for the interfaces if they can't control the generated code anyway?

ahcorde:
friendly ping @jacobperron

PR #272, opened 2020-05-22
Fix server_queue_goals.py example.

### LINKED ISSUE #271: Goal queuing action server won't allow for cancels
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Focal 20.04 ARM64 
- Installation type:
  - Binaries
- Version or commit hash:
  - [d51db0c680224c9ad7aeafc04d2cfdffa511c813](https://github.com/ros2/examples/tree/d51db0c680224c9ad7aeafc04d2cfdffa511c813)
- Client library (if applicable):
  - `rclpy`

#### Steps to reproduce issue

Start the server:

```sh
ros2 run examples_rclpy_minimal_action_server server_queue_goals
```

Send a goal and try to cancel, e.g. using:

```sh
ros2 run examples_rclpy_minimal_action_client client_cancel
```

#### Expected behavior
Goal gets accepted and later cancelled when requested.

#### Actual behavior
Goal cannot be cancelled once started.

#### Additional information
The issue here is that this _one goal at a time_ behavior is achieved by means of a single threaded executor. This doesn't allow concurrent goal executions, but it also prevents goal cancel callbacks to be served. It turns out it's not particularly easy to tell `rclpy.action` to defer a goal handle for later execution. CC @jacobperron for valuable input.

hidmic:
Fixes #271. Tested locally.

### DIFF
--- a/rclpy/actions/minimal_action_server/examples_rclpy_minimal_action_server/server_queue_goals.py
+++ b/rclpy/actions/minimal_action_server/examples_rclpy_minimal_action_server/server_queue_goals.py
@@ -71,50 +71,53 @@ class MinimalActionServer(Node):
 
     def execute_callback(self, goal_handle):
         """Executes a goal."""
-        self.get_logger().info('Executing goal...')
+        try:
+            self.get_logger().info('Executing goal...')
 
-        # Append the seeds for the Fibonacci sequence
-        feedback_msg = Fibonacci.Feedback()
-        feedback_msg.sequence = [0, 1]
+            # Append the seeds for the Fibonacci sequence
+            feedback_msg = Fibonacci.Feedback()
+            feedback_msg.sequence = [0, 1]
 
-        # Start executing the action
-        for i in range(1, goal_handle.request.order):
-            if goal_handle.is_cancel_requested:
-                goal_handle.canceled()
-                self.get_logger().info('Goal canceled')
-                result = Fibonacci.Result()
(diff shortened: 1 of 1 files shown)


rclpy/actions/minimal_action_server/examples_rclpy_minimal_action_server/server_queue_goals.py
```diff
@@ -73,14 +94,24 @@ def execute_callback(self, goal_handle):
 
             # Sleep for demonstration purposes
             time.sleep(1)
+        else:
```
jacobperron, line 97: It's just preference, but I dislike for-else blocks, because I find them hard to parse at first glance. I would also expect inexperienced Python users to be confused too.

What if instead we wrapped the logic for starting the next goal into a function and call that just before returning the result?

hidmic: See https://github.com/ros2/examples/pull/272/commits/e4ba2ee9848ebe15ad7e75363ca8fe4629e447fc. I migrated to a try-finally block, to ensure functionality doesn't break if `execute_callback` were to `raise` (in the most general case).

jacobperron: Works for me, thanks.

PR #298, opened 2021-01-06
Multithreaded executor with servers example

96moustafa:
Is it possible to add an example that explains how to implement concurrent service callbacks in a ROS2 node?


clalancette (changes requested):
From a brief glance, this looks very similar to the code in https://github.com/ros2/examples/blob/master/rclcpp/executors/multithreaded_executor/multithreaded_executor.cpp.  Can you explain how this is different?

Even if it is different, I'd want to put the changed code into the same package; it just doesn't seem like we need new packages for all of this.

PR #300, opened 2021-01-21
Add type adaption example

audrow:
This PR shows how type adaption could be for publishers and subscribers.

This PR relies on https://github.com/ros2/rclcpp/pull/1557.

<hr>

This is part of the Type Masquerading project.


rclcpp/topics/minimal_subscriber/member_function_type_masquerading.cpp
```diff
@@ -0,0 +1,48 @@
...
+public:
+  MinimalSubscriber()
+  : Node("minimal_subscriber")
+  {
+    subscription_ = this->create_subscription<std::string>(
+      "topic", 10, std::bind(&MinimalSubscriber::topic_callback, this, _1));
+  }
+
+private:
+  void topic_callback(const std::shared_ptr<std::string> msg) const
```
wjwwood, line 35: I think as part of this work we may need to rethink what callback signatures we support.

For instance, I think either of these makes more sense than this signature (though it might be the right signature atm):

- `std::shared_ptr<const std::string> msg`
- `const std::shared_ptr<const std::string> & msg`
  - arguably just this one based on https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-const

In both cases we make clear it's the string, not the shared pointer that needs to be unchanged with in this function (because we might share it with other functions).

If the user wants to own a non-const copy of the message data (so they can mutate it) then we can let them make their own copy and we can give the const reference with this signature:

- `const std::string & msg`

If we're nice we could have some syntactic sugar to make the copy for them:

- `std::string msg` or
- `std::unique_ptr<std::string> msg`
  - they could upgrade this to a shared_ptr if they like

----

Nothing to do in this example, I just wanted to mention it while I was thinking about it. If we do change things we'd need to reflect them in this example, but also all the others. And we'd have to do a tick-tock on it (deprecation cycle).

rclcpp/topics/minimal_publisher/member_function_type_masquerading.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+private:
+  void timer_callback()
+  {
+    std::string message = "Hello, world! " + std::to_string(count_++);
+    RCLCPP_INFO(this->get_logger(), "Publishing: '%s'", message.c_str());
+    publisher_->publish(message);
+  }
+  rclcpp::TimerBase::SharedPtr timer_;
+  rclcpp::Publisher<std::string>::SharedPtr publisher_;
```
wjwwood, line 47: I know we discussed this in the past, but another option here is to be more explicilt and say what `std::string` is a stand in for.

That would have the benefit of being more explicit to read and also would potentially allow us to have `std::string` map to more than one ROS type, for example.

Ideas for how that might look:

```c++
rclcpp::Publisher<std::string, std_msgs::msg::String>::SharedPtr publisher_;
```

Or the reverse order might make sense:

```c++
rclcpp::Publisher<std_msgs::msg::String, std::string>::SharedPtr publisher_;
```

But that tells me this is actually kind of ambiguous, so maybe some machinery to make it clear (though less concise):

```c++
rclcpp::Publisher<Masquerade<std::string>::as<std_msgs::msg::String>>::SharedPtr publisher_;
```

Again the order of the type could be reversed and kind of make sense? Maybe this is clearer:

```c++
rclcpp::Publisher<Use<std::string>::InsteadOf<std_msgs::msg::String>>::SharedPtr publisher_;
```

I dunno, the naming can be tweaked, but in general having some syntax to make it clear what's happening might help. We could always support both the concise and verbose options too I guess.

wjwwood: Talking with @audrow, we thought:

- Call it something else? Masquerade is a weird word, might be harder for non-English (non-French) speakers.
  - Facade (also French, but more common) makes more sense? It's shorter and the metaphor for Facade is better since it has roots in architecture.
- How should it be stated? `Facade<std::string>::instead_of<std_msgs::msg::String>` or `Facade<std_msgs::msg::String>::instead_use<std::string>`?
  - We think `Facade<std::string>::instead_of<std_msgs::msg::String>` because the "facade" is the `std::string`, and it reads better. Also the one we're "using" is farther to the left and so you see it first.
- Should we support more than one type mapping per user type?
  - It would be useful, but makes the implicit declaration (`Publisher<std::string>`) more difficult because it would only work if there's one ROS type that `std::string` could represent.

How might it work:

```c++
template<class From, class To>
struct TypeMapping
{
  // ...
};

template<class FacadeType>
struct TypeFacade {
  template<class InterfaceType>
  using instead_of = TypeMapping<FacadeType, InterfaceType>;
};

// you can be explicit like this:
using MsgT = rclcpp::TypeFacade<std::string>::instead_of<std_msgs::msg::String>;
rclcpp::Publisher<MsgT>::SharedPtr publisher_;
// ^ you would use this if you want it to be clear what is happening
//   (maybe examples or depending on the preferences of the developer), or
//   you might use it when `std::string` could map to more than one ROS type
//   and you need to select which one to use.

// We could have a short form of this that takes two arguments:
rclcpp::Publisher<TypeMapping<std::string, std_msgs::msg::String>>::SharedPtr publisher_;
// Or this, but it might be prohibitively complicated to implement
// since there are multiple template args already for Publisher/Subscription:
rclcpp::Publisher<std::string, std_msgs::msg::String>::SharedPtr publisher_;

// or you can be implicit:
rclcpp::Publisher<std::string>::SharedPtr publisher_;
// ^ If std::string has multiple types it could represent then this has to error.
```

gavanderhoorn: > Call it something else? Masquerade is a weird word, might be harder for non-English (non-French) speakers.

the technique is called *type masquerading*. It seems a bit artificial to try and find a different name for it in ROS 2, especially since there has been major effort in its design to use standard(ised) technologies and approaches in the overall design. Unless there is a really good reason not to use the established name for this, I would recommend against it.

Additionally, *facading* is not a verb, while masquerading is. Minor, but makes it easier to discuss it.

The code sample reads a bit "forced" as well: 

```c++
Facade<...>::instead_of<...>
```

is not nearly as easy to grok as:

```c++
rclcpp::Publisher<Masquerade<...>::as<...>>::SharedPtr publisher_;
```

which communicates the *use-this-instead-of-that* intent very nicely.

wjwwood: Thanks, we'll take that into consideration, and we're planning on making a REP and listing this in the alternative section (which ever one we do not select for the name).

> the technique is called type masquerading. It seems a bit artificial to try and find a different name for it in ROS 2,

When @audrow mentioned to me that the name might be confusing and we should consider alternatives, I was open to it mostly because AFAIK I had just invented the term `type masquerading` years ago. It's possible I didn't and instead I heard it from some ROS 1 stuff, but recently when I was spinning back up on this topic and looking at ROS 1's documentation on this, they don't call it that at all:

http://wiki.ros.org/roscpp/Overview/MessagesSerializationAndAdaptingTypes#Adapting_C.2B-.2B-_Types

It's referred to as "adapting C++ types", at least in the documentation.

I remember taking inspiration from IP Masquerading, which is a somewhat related concept to do with NAT routing, but other than that I'm not aware of a broader CS technique about `type masquerading`.

After talking about it `facade` was interesting for the reasons I noted above.

That all being said I actually don't mind what's called, and I take the point that you think `Masquerade<...>::as<...>` is clearer.

Just out of curiosity which do you think is more correct @gavanderhoorn?

```c++
Masquerade<std_msgs::msg::String>::as<std::string>
// or
Masquerade<std::string>::as<std_msgs::msg::String>
```

> Additionally, facading is not a verb (it is a band apparently :) ), while masquerading is. Minor, but makes it easier to discuss it.

I suppose so, but I was thinking of "a Type Facade", and in the code `TypeFacade<X>` is more correct in an object oriented sense, as `Masquerade<X>` reads a both a noun and a verb and as a noun it doesn't make sense to me. Again, it doesn't really matter to me.

gavanderhoorn: >  > the technique is called type masquerading. It seems a bit artificial to try and find a different name for it in ROS 2,
> 
> When @audrow mentioned to me that the name might be confusing and we should consider alternatives, I was open to it mostly because AFAIK I had just invented the term type masquerading years ago.
> 
> I remember taking inspiration from IP Masquerading, which is a somewhat related concept to do with NAT routing, but other than that I'm not aware of a broader CS technique about `type masquerading`.

So I think you may be right actually. I've heard the term being used around me a lot, but perhaps that's in a different context. IP masquerading is indeed a thing, as is a masquerade attack (pretending to be an authorised user). There are some other contexts in which the term is used in a similar fashion, but I can't find an official(-ish) reference for you specifically dealing with *types* in the sense that we're discussing here.

So to me, it's completely clear what is meant by *type masquerading*. I'd also assume that for anyone familiar with facades, proxies, wrappers or similar constructs, it would not difficult to infer the meaning.

But if you don't know the connection between mask and masquerade, then it could be difficult to figure out.

That said though: I believe the term does work for the intended purpose.

> Just out of curiosity which do you think is more correct @gavanderhoorn?
> 
> ```
> Masquerade<std_msgs::msg::String>::as<std::string>
> // or
> Masquerade<std::string>::as<std_msgs::msg::String>
> ```

so if "to masquarade" is intended to mean: *pretend to be something/someone else*, then only the second option would seem to be correct, and would mean: take `std::string`, and anywhere you find me using a `std::string` (in specific contexts) pretend I wrote `std_msgs::msg::String` instead.

Which seems to be what you're after, right?

---

Thinking about it some more: `Pretend<..>::is_a<..>` could be an alternative.

wjwwood: Again, we'll take that into consideration, but if we continue this discussion maybe on the new draft pr for the rep is a better place to discuss it:

https://github.com/ros-infrastructure/rep/pull/303

We have a section there where we are discussing naming and alternatives considered.

rclcpp/topics/minimal_publisher/member_function_with_type_adapter.cpp
```diff
@@ -0,0 +1,88 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
wjwwood, line 1:
```suggestion
// Copyright 2021 Open Source Robotics Foundation, Inc.
```

Or if you like:

```suggestion
// Copyright 2016-2021 Open Source Robotics Foundation, Inc.
```

rclcpp/topics/minimal_publisher/member_function_with_type_adapter.cpp
```diff
@@ -0,0 +1,88 @@
...
+    const ros_message_type & source,
+    custom_type & destination)
+  {
+    destination = source.data;
+  }
+};
+
+
+/* This example creates a subclass of Node and uses std::bind() to register a
+ * member function as a callback from the timer. */
```
wjwwood, line 55: Maybe update this

rclcpp/topics/minimal_subscriber/member_function_with_type_adapter.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "rclcpp/type_adapter.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+#include "std_msgs/msg/string.hpp"
+
+using std::placeholders::_1;
+
+template<>
+struct rclcpp::TypeAdapter<std::string, std_msgs::msg::String>
```
wjwwood, line 27: Can we have a comment here, and in the publisher equivalent, that normally a specialization like this would be in a header and re-used by the publisher and subscription rather than copy-pasted like this? I still think it's good as-is because it keeps us from sharing files between the examples and makes them more "self-contained".

rclcpp/topics/minimal_subscriber/member_function_with_type_adapter.cpp
```diff
@@ -0,0 +1,79 @@
...
+  void
+  convert_to_custom(
+    const ros_message_type & source,
+    custom_type & destination)
+  {
+    destination = source.data;
+  }
+};
+
+
```
wjwwood, line 52: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/topics/minimal_subscriber/member_function_with_type_adapter.cpp
```diff
@@ -0,0 +1,79 @@
...
+public:
+  MinimalSubscriber()
+  : Node("minimal_subscriber")
+  {
+    subscription_ = this->create_subscription<MyAdaptedType>(
+      "topic", 10, std::bind(&MinimalSubscriber::topic_callback, this, _1));
+  }
+
+private:
+  void topic_callback(const std::shared_ptr<std::string> msg) const
```
wjwwood, line 66: Let's go ahead and update this to a `const std::string & msg`, since I think that's what I would recommend these days unless you really needed a shared pointer or unique pointer for some reason (which we don't here).

rclcpp/topics/minimal_subscriber/member_function_with_type_adapter.cpp
```diff
@@ -0,0 +1,79 @@
...
+  {
+    RCLCPP_INFO(this->get_logger(), "I heard: '%s'", (*msg).c_str());
```
wjwwood, line 68: If the previous comment is addressed this won't matter, but in the future (or if we don't do the other comment), this should be:

```suggestion
    RCLCPP_INFO(this->get_logger(), "I heard: '%s'", msg->c_str());
```

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14610/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9365/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12276/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14744/)

Issue #312, opened 2021-04-27
publisher_member_function_with_unique_network_flow_endpoints leads to coredump


fujitatomoya:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - source build
- Version or commit hash:
  - https://github.com/ros2/ros2/commit/2c3dcb37040365ee5f09eede264cd05910134076
- DDS implementation:
  - cyclonedds
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
root@134f29e8f25f:~/ros2_ws/colcon_ws# ros2 run examples_rclcpp_minimal_publisher publisher_member_function_with_unique_network_flow_endpoints
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  error obtaining network flows of publisher: rmw_publisher_get_network_flow_endpoints not implemented, at /root/ros2_ws/colcon_ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_get_network_flow_endpoints.cpp:31, at /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl/src/rcl/network_flow_endpoints.c:83, also error cleaning up network flow array: network_flow_endpoint_array->allocator is null, at /root/ros2_ws/colcon_ws/src/ros2/rmw/rmw/src/network_flow_endpoint_array.c:81: error not set
```

#### Expected behavior

example should not get crashed even if rmw implementation does not support `rmw_publisher_get_network_flow_endpoints`

#### Actual behavior

publisher_member_function_with_unique_network_flow_endpoints coredump

#### Additional information

- rmw_fastdds_cpp supports `rmw_publisher_get_network_flow_endpoints` so this problem cannot be observed.

fujitatomoya:
@anamud friendly ping, since you were original author.

anamud:
@fujitatomoya Thanks for spotting the issue and submitting a good solution. I have reviewed the patch and suggested a minor change.

endre90:
Hi, 

I can't find much info on this type of pubs/subs, so I thought I might just ask the question here. Could somebody elaborate on what does having unique network flow endpoints mean in DDS and what are the benefits compared to nodes that don't use this option? Either that or maybe point me to a link where this is explained?

Thanks in advance,
Endre

anamud:
[...]

PR #313, opened 2021-04-27
add try&catch statement to unique network flow publisher example.

fujitatomoya:
address https://github.com/ros2/examples/issues/312

Signed-off-by: Tomoya Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
+++ b/rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
@@ -43,7 +43,6 @@ public:
 
       // Create publisher without unique network flow endpoints
       // Unique network flow endpoints are disabled in default options
-      auto options_2 = rclcpp::PublisherOptions();
       publisher_2_ = this->create_publisher<std_msgs::msg::String>("topic_2", 10);
       timer_2_ = this->create_wall_timer(
         1000ms, std::bind(&MinimalPublisherWithUniqueNetworkFlowEndpoints::timer_2_callback, this));
--- a/rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
+++ b/rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
@@ -31,22 +31,22 @@ public:
   MinimalPublisherWithUniqueNetworkFlowEndpoints()
   : Node("minimal_publisher_with_unique_network_flow_endpoints"), count_1_(0), count_2_(0)
   {
+    // Create publisher with unique network flow endpoints
+    // Enable unique network flow endpoints via options
+    auto options_1 = rclcpp::PublisherOptions();
+    options_1.require_unique_network_flow_endpoints =
+      RMW_UNIQUE_NETWORK_FLOW_ENDPOINTS_OPTIONALLY_REQUIRED;
+    publisher_1_ = this->create_publisher<std_msgs::msg::String>("topic_1", 10, options_1);
+    timer_1_ = this->create_wall_timer(
+      500ms, std::bind(&MinimalPublisherWithUniqueNetworkFlowEndpoints::timer_1_callback, this));
+
+    // Create publisher without unique network flow endpoints
+    // Unique network flow endpoints are disabled in default options
+    publisher_2_ = this->create_publisher<std_msgs::msg::String>("topic_2", 10);
+    timer_2_ = this->create_wall_timer(
+      1000ms, std::bind(&MinimalPublisherWithUniqueNetworkFlowEndpoints::timer_2_callback, this));
+
     try {
-      // Create publisher with unique network flow endpoints
-      // Enable unique network flow endpoints via options
-      auto options_1 = rclcpp::PublisherOptions();
-      options_1.require_unique_network_flow_endpoints =
-        RMW_UNIQUE_NETWORK_FLOW_ENDPOINTS_OPTIONALLY_REQUIRED;
-      publisher_1_ = this->create_publisher<std_msgs::msg::String>("topic_1", 10, options_1);
-      timer_1_ = this->create_wall_timer(
-        500ms, std::bind(&MinimalPublisherWithUniqueNetworkFlowEndpoints::timer_1_callback, this));
-
-      // Create publisher without unique network flow endpoints
-      // Unique network flow endpoints are disabled in default options
-      publisher_2_ = this->create_publisher<std_msgs::msg::String>("topic_2", 10);
-      timer_2_ = this->create_wall_timer(
-        1000ms, std::bind(&MinimalPublisherWithUniqueNetworkFlowEndpoints::timer_2_callback, this));
-
       // Get network flow endpoints
       auto network_flow_endpoints_1 = publisher_1_->get_network_flow_endpoints();
       auto network_flow_endpoints_2 = publisher_2_->get_network_flow_endpoints();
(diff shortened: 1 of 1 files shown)


fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14487/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9247/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12153/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14587/)

anamud (changes requested):
Thanks for submitting a good solution. Since the error is due to missing implementation of `get_network_flow_endpoints()`, I suggest to narrow the scope of the `try` block and start it just before the line with comment "Get network flow endpoints" as shown below.

```c++
   try { 
      // Get network flow endpoints
      auto network_flow_endpoints_1 = publisher_1_->get_network_flow_endpoints();
      auto network_flow_endpoints_2 = publisher_2_->get_network_flow_endpoints();

      // Print network flow endpoints
      print_network_flow_endpoints(network_flow_endpoints_1);
      print_network_flow_endpoints(network_flow_endpoints_2);
    } catch (const rclcpp::exceptions::RCLError & e) {
      RCLCPP_ERROR(
        this->get_logger(),
        "Error: %s",
        e.what());
    }
```

fujitatomoya:
> I suggest to narrow the scope of the try block and start it just before the line with comment "Get network flow endpoints" as shown below.

yeah, i thought that would be good enough as example.
but `create_publisher` also possibly generates exception `RCLError` so i think the current code would be better to be in principle.

anamud:
> yeah, i thought that would be good enough as example.
> but `create_publisher` also possibly generates exception `RCLError` so i think the current code would be better to be in principle.

I agree with you reasoning. A larger try block can catch other errors. This change looks good to me and can be merged.

anamud (approved):
LGTM.

rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
```diff
@@ -31,29 +31,36 @@ class MinimalPublisherWithUniqueNetworkFlowEndpoints : public rclcpp::Node
...
 
-    // Create publisher without unique network flow endpoints
-    // Unique network flow endpoints are disabled in default options
-    auto options_2 = rclcpp::PublisherOptions();
-    publisher_2_ = this->create_publisher<std_msgs::msg::String>("topic_2", 10);
-    timer_2_ = this->create_wall_timer(
-      1000ms, std::bind(&MinimalPublisherWithUniqueNetworkFlowEndpoints::timer_2_callback, this));
+      // Create publisher without unique network flow endpoints
+      // Unique network flow endpoints are disabled in default options
+      auto options_2 = rclcpp::PublisherOptions();
```
clalancette, line 46: Why are we creating this `options_2` variable at all?  It is never used anywhere (as far as I can tell).

rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
```diff
@@ -31,29 +31,36 @@ class MinimalPublisherWithUniqueNetworkFlowEndpoints : public rclcpp::Node
...
-    print_network_flow_endpoints(network_flow_endpoints_2);
+      // Print network flow endpoints
+      print_network_flow_endpoints(network_flow_endpoints_1);
+      print_network_flow_endpoints(network_flow_endpoints_2);
+    } catch (const rclcpp::exceptions::RCLError & e) {
+      RCLCPP_ERROR(
+        this->get_logger(),
+        "Error: %s",
+        e.what());
+    }
```
clalancette, line 63: The problem with this is that now the constructor will "succeed", but not actually create all of the resources it should.

My suggestion is to actually not do this `try/except` here, but rather do it in main.  That way if we fail, we can print the error message and quit the program with a non-zero exit code.

clalancette (changes requested):

rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
fujitatomoya: correct, this will be removed.

rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
fujitatomoya: if that is supposed to mean the publisher, constructor will succeed and all of the resource will be created. but it fails on `get_network_flow_endpoints`. so there would be no problem? btw, i am also good to go with `moving try&catch statement in main`, if that is more comfortable example for user.

clalancette: > if that is supposed to mean the publisher, constructor will succeed and all of the resource will be created. but it fails on `get_network_flow_endpoints`. so there would be no problem?

Oh, I see.  Well, I think the problem here is that since we are catching the generic `RCLError`, many errors (including those creating the publisher or timer) might be caught here.  So with that said, I think there are two ways we can go here:

1. Move the try/catch to main.  That will catch all errors from the constructor.
2. Just put the try/catch around `get_network_flow_endpoints()`, since that is the error we are worried about here.

I'd personally lean towards the second one, but the first one is OK too.

fujitatomoya: > I'd personally lean towards the second one, but the first one is OK too.

okay, i am good to go with either way.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14647/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9400/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12311/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14791/)

rclcpp/topics/minimal_publisher/member_function_with_unique_network_flow_endpoints.cpp
```diff
@@ -42,18 +42,24 @@ class MinimalPublisherWithUniqueNetworkFlowEndpoints : public rclcpp::Node
...
-    // Print network flow endpoints
-    print_network_flow_endpoints(network_flow_endpoints_1);
-    print_network_flow_endpoints(network_flow_endpoints_2);
+      // Print network flow endpoints
+      print_network_flow_endpoints(network_flow_endpoints_1);
+      print_network_flow_endpoints(network_flow_endpoints_2);
+    } catch (const rclcpp::exceptions::RCLError & e) {
+      RCLCPP_ERROR(
+        this->get_logger(),
+        "Error: %s",
```
clalancette, line 60: Two more minor things here.  The first is that I think we should have a comment block at the top of the `try/catch` sequence explaining that we expect this to fail in some situations, and that is expected.

The second is that we should probably not print "Error" here, since it is not a hard error.  Maybe "Warning" or even "Info" here would be better, since this is an allowed situation.

fujitatomoya: agree, addressed with https://github.com/ros2/examples/pull/313/commits/0bd84d886660d8ad72d272fbfa90662811aeeb5e

anamud (approved):
The added comment is excellent since it captures the exact reason for the expected exception. Minor suggestion: I would have preferred to use RCLCPP_WARN instead of RCLCPP_INFO.

clalancette (approved):
This looks good to me now with green CI.  Thanks for the iterations.

fujitatomoya:
@clalancette could you do me a favor? i don't have access permission to this repo.

clalancette:
> @clalancette could you do me a favor? i don't have access permission to this repo.

Sure, sounds good.  I'll go ahead and merge this.

PR #315, opened 2021-06-25
Add wait set examples

carlossvg:
This PR adds examples for how to use the new rclcpp::WaitSet classes. It is an extension of the existing PR https://github.com/ros2/examples/pull/262 created by @wjwwood. 

This PR includes the following additional examples:
 * `wait_set_subscriber.cpp` uses a `rclcpp::WaitSet` to wait and poll for data
 * `static_wait_set_subscriber.cpp` uses a `rclcpp::StaticWaitSet` to wait and poll for data
 * `time_triggered_wait_set_subscriber.cpp` uses a `rclcpp::Waitset` and a timer to poll for data
   periodically
* `wait_set.cpp`: Simple example showing how to use the default wait-set with a dynamic
  storage policy and a sequential (no thread-safe) synchronization policy.
* `static_wait_set.cpp`: Simple example showing how to use the static wait-set with a static
  storage policy.
* `thread_safe_wait_set.cpp`: Simple example showing how to use the thread-safe wait-set with a
  thread-safe synchronization policy.
* `wait_set_topics_and_timer.cpp`: Simple example using multiple subscriptions,
  publishers, and a timer.
* `wait_set_random_order.cpp`: An example showing user-defined
  data handling and a random publisher. `executor_random_order.cpp` run the same node logic
  using `SingleThreadedExecutor` to compare the data handling order.
 * `wait_set_and_executor_composition.cpp`: An example showing how to combine a  
  `SingleThreadedExecutor` and a wait-set.
* `wait_set_topics_with_different_rate.cpp`: An example showing how to use a custom trigger
  condition to handle topics with different topic rates.

Some of the examples may be moved to ros2/tutorials or ros2/demos. 

Connects to ros2/rclcpp#1047


carlossvg:
@wjwwood I will add more comments, add a readme and probably rename the examples. Could briefly you take a look and see if you miss an example or a specific usage we want to show?

rclcpp/minimal_subscriber/waitset.cpp
```diff
@@ -0,0 +1,92 @@
...
+  {
+    return subscription_;
+  }
+
+  void run()
+  {
+    rclcpp::WaitSet wait_set;
+    wait_set.add_subscription(subscription_);
+    while (rclcpp::ok()) {
+      const auto wait_result = wait_set.wait(std::chrono::milliseconds(500));
```
carlossvg, line 45: A timeout was necessary here in order to return on shutdown. Otherwise it gets blocked here.

rclcpp/wait_set/wait_set_random_order.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+      // Handle all the messages when all subscriptions have data
+      if (sub1_has_data && sub2_has_data && sub3_has_data) {
+        std_msgs::msg::String msg;
+        rclcpp::MessageInfo msg_info;
+
+        // the messages are taken and handled in a user-defined order
+        if (subscriptions.at(0)->take(msg, msg_info)) {
+          std::shared_ptr<void> type_erased_msg = std::make_shared<std_msgs::msg::String>(msg);
+          subscriptions.at(0)->handle_message(type_erased_msg, msg_info);
```
carlossvg, line 63: @wjwwood I'm not sure if this the the best approach or it's preferred to use `take_type_erased` directly.

carlossvg: I mean using this approach instead:

```cpp
std::shared_ptr<void> msg = sub1->create_message();
sub1->take_type_erased(msg.get(), msg_info);
subscriptions.at(0)->handle_message(msg, msg_info);
```

wjwwood: The `create_message()` method will use the allocator that was given when creating the subscription. I don't think that's important in this case.

I think your existing code makes sense.

carlossvg:
@wjwwood This PR should  be ready to review.

rclcpp/wait_set/thread_safe_wait_set.cpp
```diff
@@ -0,0 +1,112 @@
...
+  auto node = std::make_shared<rclcpp::Node>("wait_set_example_node");
+
+  auto do_nothing = [](std_msgs::msg::String::UniquePtr) {assert(false);};
+  auto sub1 = node->create_subscription<std_msgs::msg::String>("~/chatter", 10, do_nothing);
+  auto sub2 = node->create_subscription<std_msgs::msg::String>("~/chatter", 10, do_nothing);
+  std::vector<decltype(sub1)> sub_vector = {sub1, sub2};
+  auto guard_condition1 = std::make_shared<rclcpp::GuardCondition>();
+  auto guard_condition2 = std::make_shared<rclcpp::GuardCondition>();
+
+  // FIXME: removing sub if it was added in the ctor leads to a failure
```
carlossvg, line 35: @wjwwood I was getting an exception here when adding the subscription in the ctor. I guess this will be gone once https://github.com/ros2/rclcpp/pull/1700 is merged.

rclcpp/wait_set/README.md
```diff
@@ -0,0 +1,23 @@
...
+* `wait_set_topics_and_timer.cpp`: Simple example using multiple subscriptions,
+  publishers, and a timer.
+* `wait_set_random_order.cpp`: An example showing user-defined
+  data handling and a random publisher. `executor_random_order.cpp` run the same node logic
+  using `SingleThreadedExecutor` to compare the data handling order.
+* `wait_set_and_executor_composition.cpp`: An example showing how to combine a  
+  `SingleThreadedExecutor` and a wait-set.
+* `wait_set_topics_with_different_rate.cpp`: An example showing how to use a custom trigger
+  condition to handle topics with different topic rates.
+  and-paste into your own projects.
```
wjwwood, line 23: This line fragment looks out of place?

rclcpp/wait_set/README.md
```diff
@@ -0,0 +1,23 @@
...
+  storage policy and a sequential (no thread-safe) synchronization policy.
+* `static_wait_set.cpp`: Simple example showing how to use the static wait-set with a static
+  storage policy.
+* `thread_safe_wait_set.cpp`: Simple example showing how to use the thread-safe wait-set with a
+  thread-safe synchronization policy.
+* `wait_set_topics_and_timer.cpp`: Simple example using multiple subscriptions,
+  publishers, and a timer.
+* `wait_set_random_order.cpp`: An example showing user-defined
+  data handling and a random publisher. `executor_random_order.cpp` run the same node logic
+  using `SingleThreadedExecutor` to compare the data handling order.
```
wjwwood, line 18: This pair of executables looks more like a demo or maybe a tutorial. Remember the purpose of examples are to be copy-paste starting points, not to show that a feature works.

I think we already talked about this, and we agreed to take them here and move them later? If that's the case, then nothing to do here, but I wanted to point it out again.

rclcpp/wait_set/random_listener.hpp
```diff
@@ -0,0 +1,50 @@
...
+#ifndef RCLCPP__WAIT_SET__RANDOM_LISTENER_HPP_
+#define RCLCPP__WAIT_SET__RANDOM_LISTENER_HPP_
+
+#include <vector>
+#include <random>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+class Listener : public rclcpp::Node
```
wjwwood, line 24: Shouldn't this be:

```suggestion
class RandomListener : public rclcpp::Node
```

rclcpp/wait_set/executor_random_order.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+  rclcpp::init(argc, argv);
+
+  // We run the talker and the listener in different threads to simulate inter-process
+  // communications.
+  // Note the order of execution would be deterministic if both nodes are spun in the same
+  // executor (A, B, C). This is because the publishing happens always before the subscription
+  // handling and the executor handles the messages in the order in which the subscriptions were
+  // created. Using different threads the handling order depends on the message arrival and
+  // type of messages available.
```
wjwwood, line 39: You might want to put these in different context's too. In the future we may make intra-process communication the default, and then it might change the behavior of this example. If you want to emulate being in separate processes, I'd recommend putting them into different contexts as well.

rclcpp/wait_set/static_wait_set.cpp
```diff
@@ -0,0 +1,95 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include <vector>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+int main(int argc, char * argv[])
```
wjwwood, line 21: Maybe this examples needs a comment explaining it like the others have.

wjwwood: Also, this example also falls into that "not really an example" category, because it almost more of a demo showing each item works or a test.

I was thinking this would be more like `wait_set.cpp` but just with a static wait set.

rclcpp/wait_set/thread_safe_wait_set.cpp
wjwwood: Right this is one of the bugs being fixed.

For now you could add it after construction and it would work. Which is what you've done 👍

rclcpp/wait_set/thread_safe_wait_set.cpp
```diff
@@ -0,0 +1,112 @@
...
+// limitations under the License.
+
+#include <cassert>
+#include <memory>
+#include <vector>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+int main(int argc, char * argv[])
```
wjwwood, line 22: This example is showing off the dynamic nature of the wait set, but not really anything about the thread-safe part. Maybe we should remove, or rewrite, these examples that are more like tests.

rclcpp/wait_set/wait_set.cpp
```diff
@@ -0,0 +1,112 @@
...
+// limitations under the License.
+
+#include <cassert>
+#include <memory>
+#include <vector>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+
```
wjwwood, line 22: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/wait_set/wait_set_and_executor_composition.cpp
```diff
@@ -0,0 +1,129 @@
...
+        if (listener->get_subscription()->take(msg, msg_info)) {
+          RCLCPP_INFO(listener->get_logger(), "I heard: '%s' (wait-set)", msg.data.c_str());
+        }
+      }
+    } else if (wait_result.kind() == rclcpp::WaitResultKind::Timeout) {
+      if (rclcpp::ok()) {
+        RCLCPP_ERROR(listener->get_logger(), "Wait-set failed with timeout");
+      }
+    }
+  }
```
wjwwood, line 124: In my opinion, for this example this code should be in a thread owned by the class. The whole idea is that the node is self-contained without a main function.

rclcpp/wait_set/wait_set_topics_and_timer.cpp
```diff
@@ -0,0 +1,108 @@
...
+// limitations under the License.
+
+#include <memory>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+using namespace std::chrono_literals;
+
+/* This example creates a node with three publishers, three subscriptions and one-off timer. The
```
wjwwood, line 22:
```suggestion
/* This example creates a node with three publishers, three subscriptions and a one-off timer. The
```

?

rclcpp/wait_set/wait_set_and_executor_composition.cpp
wjwwood: In fact, it should probably just register itself as a component and not have a main function at all.

rclcpp/wait_set/wait_set_random_order.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * publish the topics A, B, C in random order each time. The order in which the messages are
+ * handled is defined deterministically by the user in the code using a wait-set based loop.
+ * That is, in this example we always take and process the data in the same order  A, B, C
+ * regardless of the arrival order. */
+
+int32_t main(const int32_t argc, char ** const argv)
+{
+  rclcpp::init(argc, argv);
+
+  auto listener = std::make_shared<Listener>();
```
wjwwood, line 35: I think you still need to spin the listener, even if you're waiting on the subscriptions with a wait set.

carlossvg: Is this required as a best practice, like for instance to be consistent with the other examples or is there an specific reason to spin the node in this case?

rclcpp/wait_set/wait_set_and_executor_composition.cpp
carlossvg: Makes sense. Do I create a separate `.hpp` file for this and then a separate main in `.cpp` file?

rclcpp/wait_set/thread_safe_wait_set.cpp
carlossvg: I think it is worth to have one example for the thread safe wait-set at least to make it more visible for the users. We could remove all the dynamic add/remove stuff and simplify it just to show how it is used. Another option would be to create a more specific example, like sharing data between threads. It won't be too impressive probably but it could be a good minimal example where it makes sense to use it.

rclcpp/wait_set/README.md
carlossvg: I agree. I think this one should go on demos or tutorials. Is this the only executable we should move?

rclcpp/wait_set/executor_random_order.cpp
carlossvg: I realized it's not really about inter-process communication but it's more about executing nodes in the same executor. I removed this comment.

rclcpp/wait_set/static_wait_set.cpp
carlossvg: I created a static minimal subscriber version in the `minimal_subscriber` package.

rclcpp/wait_set/wait_set.cpp
carlossvg: done

rclcpp/wait_set/random_listener.hpp
carlossvg: done

rclcpp/wait_set/README.md
carlossvg: removed

rclcpp/minimal_subscriber/waitset.cpp
```diff
@@ -0,0 +1,78 @@
...
+private:
+  rclcpp::Subscription<std_msgs::msg::String>::SharedPtr subscription_;
+  rclcpp::WaitSet wait_set_;
+};
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  auto node = std::make_shared<MinimalSubscriber>();
+  node->run();
```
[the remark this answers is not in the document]
carlossvg: I added the wait-set loop to a thread owned by the class as you suggested. I'm not sure if now we should add a start function or something to sync the node spinning with the loop. Otherwise, the node will start the wait-set loop when the node is constructed. 

Also, note I renamed `run()` to `spin_wait_set()` si is more consistent with .the Node logic. I can revert this change if it's confusing.

rclcpp/wait_set/wait_set_and_executor_composition.cpp
carlossvg: I created the Nodes as components following the same approach as in ros2/demos. Let me know if this makes sense.

rclcpp/wait_set/wait_set_random_order.cpp
wjwwood: I think it is a best practice, but for example, if you're using it with `use_sim_time = true` then you need to spin it to have time move forward.

wjwwood:
@Mergifyio rebase

rclcpp/wait_set/thread_safe_wait_set.cpp
wjwwood: I was thinking more like an example that initially adds some things, runs the wait set in a thread, then adds/removes items concurrently, to show that it's safe.

I mean, "using" the thread-safe wait set is really just creating one and doing everything else normally. Demonstrating or proving that it works is harder to do. I think, as trivial as it is, just having the minimal example that creates a thread-safe wait set instead of a normal is just fine.

rclcpp/minimal_subscriber/wait_set.cpp
```diff
@@ -0,0 +1,88 @@
...
+// limitations under the License.
+
+#include <iostream>
+#include <memory>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+/* This example creates a subclass of Node and uses a wait-set based loop to wait and handle
+ * messages from the subscriber */
```
wjwwood, line 22:
```suggestion
/* This example creates a subclass of Node and uses a wait-set based loop to wait on
 * a subscription to have messages available and then handles them manually without an executor */
```

rclcpp/minimal_subscriber/wait_set.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+class MinimalSubscriber : public rclcpp::Node
```
wjwwood, line 24: Similar request, let's call it `WaitSetSubscriber` and change the file name to match.

rclcpp/minimal_subscriber/wait_set_time_triggered.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include <iostream>
+#include <memory>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+using namespace std::chrono_literals;
+
+/* This example creates a subclass of Node and uses a wait-set based loop and polls periodically
```
wjwwood, line 23:
```suggestion
/* This example creates a subclass of Node and uses a wait-set based loop to periodically poll
```

rclcpp/minimal_subscriber/wait_set_time_triggered.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * for messages in a timer callback using a wait-set based loop. */
+
+class MinimalSubscriber : public rclcpp::Node
```
wjwwood, line 26: Let's call this one something like `TimeTriggeredWaitSetSubscriber`, and change the file name to match.

wjwwood (commented):
I left some additional comments, mostly minor things.

I would like to discuss (in a call if we can) the examples in `examples_rclcpp_wait_set`, just to help me understand better.

I think that if we can avoid the headers (may or may not be possible) then it would save us some trouble, but we can discuss that too.

rclcpp/minimal_subscriber/static_wait_set.cpp
```diff
@@ -0,0 +1,91 @@
...
+        RCLCPP_INFO(this->get_logger(), "I heard: '%s'", msg->data.c_str());
+      };
+
+    subscription_ = this->create_subscription<std_msgs::msg::String>(
+      "topic",
+      10,
+      subscription_callback,
+      options);
+
+    wait_set_ptr_ = std::make_shared<rclcpp::StaticWaitSet<1, 0, 0, 0, 0, 0>>(
```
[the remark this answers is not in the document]
carlossvg: Yeah, I tried that at first. The main problem was passing the subscription options with the new callback-group. Here it would be useful to have a subscription creation method which creates directly a subscription passing a callback-group and and without needing a callback. I'm not sure if this is feasible or not.

rclcpp/wait_set/README.md
```diff
@@ -0,0 +1,22 @@
+# Minimal rclcpp wait-set cookbook recipes
+
+This package contains a few different strategies for creating nodes which use `rclcpp::waitset`
+to wait and handle ROS entities, that is, subscribers, timers, clients, services, guard
+conditions and waitables.
```
wjwwood, line 5: Let's add a note about when and why you might want to create applications like this, i.e. you probably only want to do this in very specific situations, or not at all (in which case these are more demonstrations than examples you should copy-paste-edit).

rclcpp/minimal_subscriber/wait_set_time_triggered.cpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+private:
+  rclcpp::Subscription<std_msgs::msg::String>::SharedPtr subscription_;
+  rclcpp::TimerBase::SharedPtr timer_;
+  rclcpp::WaitSet wait_set_;
+  std::thread thread_;
+};
+
+int main(int argc, char * argv[])
```
wjwwood, line 91: Let's try to get rid of these main functions, use register component and use the executable created automatically for components. (if possible)

carlossvg:
@wjwwood I addressed your comments. You can review it again.

rclcpp/README.md
```diff
@@ -1,4 +1,4 @@
 # rclcpp examples
 
 This directory contains many examples of how to do common tasks with rclcpp.
-The intent is that this material will be easy to copy-and-paste into your own projects.
+The intent is that this material will be easy to copy-and-paste into your own projects.
```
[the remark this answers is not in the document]
carlossvg: It should be gone now

rclcpp/minimal_subscriber/static_wait_set.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <iostream>
+#include <memory>
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+/* This example creates a subclass of Node and uses static a wait-set based loop to wait and handle
+ * messages from the subscriber */
+
+class MinimalSubscriber : public rclcpp::Node
```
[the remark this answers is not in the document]
carlossvg: Renamed

rclcpp/minimal_subscriber/static_wait_set.cpp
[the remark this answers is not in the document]
carlossvg: Solved by using a lambda

rclcpp/minimal_subscriber/static_wait_set.cpp
```diff
@@ -0,0 +1,91 @@
...
+      };
+
+    subscription_ = this->create_subscription<std_msgs::msg::String>(
+      "topic",
+      10,
+      subscription_callback,
+      options);
+
+    wait_set_ptr_ = std::make_shared<rclcpp::StaticWaitSet<1, 0, 0, 0, 0, 0>>(
+      std::array<rclcpp::StaticWaitSet<1, 0, 0, 0, 0, 0>::SubscriptionEntry, 1>{{{subscription_}}});
```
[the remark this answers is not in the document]
carlossvg: I added the suggested alias

rclcpp/minimal_subscriber/static_wait_set.cpp
```diff
@@ -0,0 +1,91 @@
...
+          if (subscription_->take(msg, msg_info)) {
+            std::shared_ptr<void> type_erased_msg = std::make_shared<std_msgs::msg::String>(msg);
+            subscription_->handle_message(type_erased_msg, msg_info);
+          }
+        }
+      } else if (wait_result.kind() == rclcpp::WaitResultKind::Timeout) {
+        if (rclcpp::ok()) {
+          RCLCPP_WARN(this->get_logger(), "Timeout. No message received after given wait-time");
+        }
+      }
```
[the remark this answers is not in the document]
carlossvg: I replaced the if statements with a switch

rclcpp/minimal_subscriber/wait_set.cpp
carlossvg: renamed

rclcpp/minimal_subscriber/wait_set_time_triggered.cpp
carlossvg: renamed

rclcpp/wait_set/README.md
carlossvg: Added a note

rclcpp/minimal_subscriber/wait_set_time_triggered.cpp
carlossvg: Main functions are replaced with components. I was able to load these nodes in a container. Please review the CMakeLists.txt to make sure I'm not missing anything. Also, I'm not using any namespace for the component names, I hope this is fine.

rclcpp/topics/minimal_subscriber/CMakeLists.txt
```diff
@@ -32,6 +33,26 @@ ament_target_dependencies(subscriber_member_function_with_unique_network_flow_en
 add_executable(subscriber_not_composable not_composable.cpp)
 ament_target_dependencies(subscriber_not_composable rclcpp std_msgs)
 
+add_library(wait_set_subscriber SHARED wait_set_subscriber.cpp)
+ament_target_dependencies(wait_set_subscriber rclcpp rclcpp_components std_msgs)
+rclcpp_components_register_nodes(wait_set_subscriber WaitSetSubscriber)
```
wjwwood, line 38: Do you get an executable for each of these? If not, then I think you want `rclcpp_compnents_register_node()`? That way you can specify an executable name with `EXECUTABLE <name>`. For example:

https://github.com/ros2/demos/blob/3af19c93c7abd9a83caca47de0780e941faeb22b/demo_nodes_cpp/CMakeLists.txt#L82-L84

The plural one is for libraries that contain more than one node, see this presentation (search for `rclcpp_components_register_node`): https://roscon.ros.org/2019/talks/roscon2019_composablenodes.pdf

rclcpp/wait_set/README.md
wjwwood: Where did you add it? I still just see this paragraph and the list of nodes below.

rclcpp/minimal_subscriber/wait_set_time_triggered.cpp
wjwwood: I left another comment, but I think you do need to change how they work. I'll resolve this though.

rclcpp/wait_set/README.md
carlossvg: I added it in the `minimal_subscriber` README. Do I move the note here?

>The following examples `wait_set_subscriber.cpp`, `static_wait_set_subscriber.cpp` and 
`time_triggered_wait_set_subscriber.cpp` show how to use a subscription in a node using a `rclcpp` 
wait-set. This is not a common use case  in ROS 2 so this is not the recommended strategy to 
use by-default. This strategy makes sense in some specific situations, for example when the 
developer needs to have more control over callback order execution, to create custom triggering 
conditions or to use the timeouts provided by the  wait-sets.   

https://github.com/ros2/examples/pull/315/files#diff-7cfbf34595c2e8d6c230d5ee9d207ed0deca9f060b26da8334b512b9755f7b65R20-R25

rclcpp/topics/minimal_subscriber/CMakeLists.txt
carlossvg: I moved all the `cpp` files to one library and now I'm using `rclcpp_compnents_register_node` to register each component individually. I updated the `CMakeLists.txt` from wait_set package too.

rclcpp/topics/minimal_subscriber/README.md
```diff
@@ -4,11 +4,22 @@ This package contains a few different strategies for creating nodes which receiv
...
+All of these nodes do the same thing: they create a node called `minimal_subscriber` and subscribe to a topic named `topic` which is of datatype `std_msgs/String`.
 When a message arrives on that topic, the node prints it to the screen.
 We provide multiple examples of different coding styles which achieve this behavior in order to demonstrate that there are many ways to do this in ROS 2.
+
+The following examples `wait_set_subscriber.cpp`, `static_wait_set_subscriber.cpp` and 
+`time_triggered_wait_set_subscriber.cpp` show how to use a subscription in a node using a `rclcpp` 
+wait-set. This is not a common use case  in ROS 2 so this is not the recommended strategy to 
+use by-default. This strategy makes sense in some specific situations, for example when the 
+developer needs to have more control over callback order execution, to create custom triggering 
+conditions or to use the timeouts provided by the  wait-sets.
```
wjwwood, line 25: nitpick: please use one sentence per line

> - [any] Each sentence must start on a new line.
>   - Rationale: For longer paragraphs a single change in the beginning makes the diff unreadable since it carries forward through the whole paragraph.

https://docs.ros.org/en/rolling/Contributing/Code-Style-Language-Versions.html#markdown-restructured-text-docblocks

```suggestion
The following examples `wait_set_subscriber.cpp`, `static_wait_set_subscriber.cpp` and `time_triggered_wait_set_subscriber.cpp` show how to use a subscription in a node using a `rclcpp` wait-set.
This is not a common use case in ROS 2 so this is not the recommended strategy to  use by-default.
This strategy makes sense in some specific situations, for example when the developer needs to have more control over callback order execution, to create custom triggering conditions or to use the timeouts provided by the  wait-sets.   
```

wjwwood (approved):
lgtm, with one doc suggestion.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14917/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9659/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12588/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15076/)

carlossvg:
@wjwwood I added your suggestion. Let me know when I should clean the commit history.

wjwwood:
We squash on merge so you don't need to clean up commit history unless you really want to. If so, do it now.

carlossvg:
It's fine for me, we can squash as it is. Thanks for the review.

wjwwood:
The test failures for Windows are on nightly too. I think it was just resolved with https://github.com/ros2/rclcpp/pull/1726, so I'll rerun CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14926/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9668/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12600/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15086/)

wjwwood:
@carlossvg there are some Windows specific compiler warnings which we'll have to resolve before merging, but otherwise lgtm.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14930/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9672/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12604/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15094/)

wjwwood:
I fixed some of the warning, but didn't follow up yet on the rest. I can do that hopefully tonight and get this merged.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14956/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9696/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12633/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15119/)

PR #316, opened 2021-07-01
Add an example about how to use wait_for_all_acked

Barry-Xu-2018:
This commit depends on the merge of ros2/rcl#913 and ros2/rclcpp#1662.

### DIFF
--- a/rclcpp/topics/minimal_publisher/CMakeLists.txt
+++ b/rclcpp/topics/minimal_publisher/CMakeLists.txt
@@ -26,6 +26,9 @@ ament_target_dependencies(publisher_member_function_with_type_adapter rclcpp std
 add_executable(publisher_member_function_with_unique_network_flow_endpoints member_function_with_unique_network_flow_endpoints.cpp)
 ament_target_dependencies(publisher_member_function_with_unique_network_flow_endpoints rclcpp std_msgs)
 
+add_executable(publisher_member_function_with_wait_for_all_acked member_function_with_wait_for_all_acked.cpp)
+ament_target_dependencies(publisher_member_function_with_wait_for_all_acked rclcpp std_msgs)
+
 add_executable(publisher_not_composable not_composable.cpp)
 ament_target_dependencies(publisher_not_composable rclcpp std_msgs)
 
@@ -34,6 +37,7 @@ install(TARGETS
   publisher_member_function
   publisher_member_function_with_type_adapter
   publisher_member_function_with_unique_network_flow_endpoints
+  publisher_member_function_with_wait_for_all_acked
   publisher_not_composable
   DESTINATION lib/${PROJECT_NAME}
 )
--- a/rclcpp/topics/minimal_publisher/CMakeLists.txt
+++ b/rclcpp/topics/minimal_publisher/CMakeLists.txt
@@ -26,8 +26,8 @@ ament_target_dependencies(publisher_member_function_with_type_adapter rclcpp std
 add_executable(publisher_member_function_with_unique_network_flow_endpoints member_function_with_unique_network_flow_endpoints.cpp)
 ament_target_dependencies(publisher_member_function_with_unique_network_flow_endpoints rclcpp std_msgs)
 
-add_executable(publisher_member_function_with_wait_for_all_acked member_function_with_wait_for_all_acked.cpp)
-ament_target_dependencies(publisher_member_function_with_wait_for_all_acked rclcpp std_msgs)
+add_executable(publisher_wait_for_all_acked member_function_with_wait_for_all_acked.cpp)
+ament_target_dependencies(publisher_wait_for_all_acked rclcpp std_msgs)
 
 add_executable(publisher_not_composable not_composable.cpp)
 ament_target_dependencies(publisher_not_composable rclcpp std_msgs)
@@ -37,7 +37,7 @@ install(TARGETS
   publisher_member_function
   publisher_member_function_with_type_adapter
   publisher_member_function_with_unique_network_flow_endpoints
-  publisher_member_function_with_wait_for_all_acked
+  publisher_wait_for_all_acked
   publisher_not_composable
   DESTINATION lib/${PROJECT_NAME}
 )
--- a/rclcpp/topics/minimal_publisher/member_function_with_wait_for_all_acked.cpp
+++ b/rclcpp/topics/minimal_publisher/member_function_with_wait_for_all_acked.cpp
@@ -24,21 +24,69 @@ using namespace std::chrono_literals;
 
 /* This example shows how to use wait_for_all_acked for the publisher */
 
+class CustomContext : public rclcpp::Context
+{
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS(CustomContext)
+
+  bool shutdown(const std::string & reason) override
+  {
+    // Before shutdown, confirm all subscribers acknowledge sent messages
+    if (custom_shutdown_func_) {
+      custom_shutdown_func_();
+    }
+
+    return rclcpp::Context::shutdown(reason);
+  }
+
+  void set_custom_shutdown_func(std::function<void()> custom_shutdown_func)
+  {
+    custom_shutdown_func_ = custom_shutdown_func;
+  }
+
+private:
+  std::function<void()> custom_shutdown_func_;
+};
+
 class MinimalPublisher : public rclcpp::Node
 {
 public:
-  MinimalPublisher()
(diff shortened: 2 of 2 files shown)


Barry-Xu-2018:
If pre-shutdown callback is supported (https://github.com/ros2/rclcpp/issues/1706#issuecomment-872613908), I can simplify current codes.

fujitatomoya (approved):

fujitatomoya:
probably it would be nice to have only specified example to introduce `wait_for_acknowledgement` w/o `custom_shutdown_func`? but i think it is still okay, any opinions?

Barry-Xu-2018:
> probably it would be nice to have only specified example to introduce `wait_for_acknowledgement` w/o `custom_shutdown_func`? but i think it is still okay, any opinions?

Yes. 
The example want to show how to wait the confirmation of subscribers while calling shutdown.   
But there is no pre-shutdown callback interface.   
So I have to use `custom_shutdown_func` to implement it.   
About 'pre-shutdown', it was discussed in https://github.com/ros2/rclcpp/issues/1706#issuecomment-872613908.   
After this feature is implemented, we can simplify this example.

fujitatomoya:
@Barry-Xu-2018 can we also have example for `rclpy`?

Barry-Xu-2018:
> @Barry-Xu-2018 can we also have example for `rclpy`?

Yeah. I will consider adding this example.

fujitatomoya:
we can run CI once https://github.com/ros2/rclcpp/pull/1662 is merged.

Barry-Xu-2018:
Example code for python isn't added.   
It is because rclpy has this issue ros2/rclpy#532. I added comments https://github.com/ros2/rclpy/issues/532#issuecomment-972610972.

Barry-Xu-2018:
do rebase

fujitatomoya:
@ros-pull-request-builder retest this please

fujitatomoya:
@clalancette it appears that CI still does not have https://github.com/ros2/rclcpp/pull/1662, that leads this PR to build failure. Do you know how long it takes? i was thinking this is done by nightly build.

clalancette:
> @clalancette it appears that CI still does not have [ros2/rclcpp#1662](https://github.com/ros2/rclcpp/pull/1662), that leads this PR to build failure. Do you know how long it takes? i was thinking this is done by nightly build.

So for the Rpr job to pass, we need to make an explicit source and bloom release to get PRs in.  I can do one of those later today.

That said, you should be able to run a CI job from https://ci.ros2.org to verify this functionality on all platforms, and that will include the latest sources.

fujitatomoya:
CI: (just checking if it can be build with mainline since there is no test for examples)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15815/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10493/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16098/)

Barry-Xu-2018:
@fujitatomoya 

I update code to fix build warning on macOS.
Could you re-trigger CI ?

fujitatomoya:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13490/)

fujitatomoya:
@ros-pull-request-builder retest this please

fujitatomoya:
@ivanpauno @sloretz @clalancette could you help this with review?

fujitatomoya:
@ros-pull-request-builder retest this please

rclcpp/topics/minimal_publisher/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ ament_target_dependencies(publisher_member_function_with_type_adapter rclcpp std
 add_executable(publisher_member_function_with_unique_network_flow_endpoints member_function_with_unique_network_flow_endpoints.cpp)
 ament_target_dependencies(publisher_member_function_with_unique_network_flow_endpoints rclcpp std_msgs)
 
+add_executable(publisher_member_function_with_wait_for_all_acked member_function_with_wait_for_all_acked.cpp)
```
ivanpauno, line 29: can we use a shorter name for the example?

It really doesn't matter if this is using a "member function", I think we could rather use "publisher_wait_for_all_acked".

rclcpp/topics/minimal_publisher/member_function_with_wait_for_all_acked.cpp
```diff
@@ -0,0 +1,94 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include <inttypes.h>
+
+#include <chrono>
+#include <functional>
```
ivanpauno, line 17:
```suggestion
#include <chrono>
#include <cstdint>
#include <functional>
```

clalancette (approved):

rclcpp/topics/minimal_publisher/member_function_with_wait_for_all_acked.cpp
```diff
@@ -0,0 +1,94 @@
...
+    wait_timeout_(300)
+  {
+    // publisher must set reliable mode
+    publisher_ = this->create_publisher<std_msgs::msg::String>(
+      "topic",
+      rclcpp::QoS(10).reliable());
+
+    // call wait_for_all_acked before shutdown
+    using rclcpp::contexts::get_global_default_context;
+    get_global_default_context()->add_pre_shutdown_callback(
```
ivanpauno, line 43: why "pre shutdown"?
The timer might still be running, so I'm not sure if this is going to wait for messages published after the method was called.
Maybe this should be done in the destructor?
Or after shutdown? Though I'm not sure if we're going to get a "context was shutdown" error in that case.

Barry-Xu-2018: > The timer might still be running.

Yeah. This is a problem. 

> Maybe this should be done in the destructor? 
> Or after shutdown?

wait_for_all_acked() cannot be called after shutdown.  
It is because rcl layer is shutdown. I will cancel timer before calling wait_for_all_acked().

```c++
ontext::shutdown(const std::string & reason)
{
 ...

  // call each pre-shutdown callback
  {
    std::lock_guard<std::mutex> lock{pre_shutdown_callbacks_mutex_};
    for (const auto & callback : pre_shutdown_callbacks_) {
      (*callback)();
    }
  }

  // rcl shutdown
  rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
  if (RCL_RET_OK != ret) {
    rclcpp::exceptions::throw_from_rcl_error(ret);
  }
  // set shutdown reason
  shutdown_reason_ = reason;
  // call each shutdown callback
  {
    std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
    for (const auto & callback : on_shutdown_callbacks_) {
      (*callback)();
    }
  }
 ...
}

Barry-Xu-2018:
Friendly ping @ivanpauno

rclcpp/topics/minimal_publisher/member_function_with_wait_for_all_acked.cpp
ivanpauno: >  I will cancel timer before calling wait_for_all_acked().

That seems okay.

> It is because rcl layer is shutdown.

I think that methods like `wait_for_all_acked()` should continue working after shutdown.
I don't see the point of making them fail.

Barry-Xu-2018: >>It is because rcl layer is shutdown.
>
> I think that methods like wait_for_all_acked() should continue working after shutdown.
> I don't see the point of making them fail.

From my understanding, once rcl is shutdown, context is invalid. And then if call  wait_for_all_acked, failure is gotten.

ivanpauno: I understand how it works.
What I mean is that doesn't seem to be useful.

IMO, we don't need much more that notifying executors to stop in practice, I think that calls like `publish()`, `wait_for_all_ack()` or anything else should still be valid after shutdown.

But anyways, the fix you pushed seems fine, and modifying how shutdown works seems out of scope.

ivanpauno (approved):

fujitatomoya:
@ivanpauno really appreciate for helping on this, thanks 👍

PR #326, opened 2021-10-11
Reverted WaitForTopics utility usage

adityapande-1995:
Removed usage of the ``WaitForTopics`` utiity that was reverted in https://github.com/ros2/launch_ros/pull/276

Signed-off-by: Aditya Pande <aditya050995@gmail.com>

### DIFF
--- a/launch_testing/launch_testing_examples/launch_testing_examples/check_msgs_launch_test.py
+++ b/launch_testing/launch_testing_examples/launch_testing_examples/check_msgs_launch_test.py
@@ -45,7 +45,7 @@ def generate_test_description():
 
 class TestFixture(unittest.TestCase):
 
-    def test_check_if_msgs_published(self, proc_output):
+    def test_check_if_msgs_published(self):
         with WaitForTopics([('chatter', String)], timeout=5.0):
             print('Topic received messages !')
 


launch_testing/launch_testing_examples/launch_testing_examples/check_msgs_launch_test.py
```diff
@@ -39,6 +42,33 @@ def generate_test_description():
...
+    def test_check_if_msgs_published(self, proc_output):
+        rclpy.init()
+        node = DummyNode('test_node')
+        node.start_subscriber()
+        msgs_received_flag = node.msg_event_object.wait(timeout=5.0)
+        assert msgs_received_flag, 'Did not receive msgs !'
+        rclpy.shutdown()
+
+
+class DummyNode(Node):
```
jacobperron, line 54: I would call this `WaitForTopics`. 

If you prefer, we can leave the test above unchanged and just copy the entire `WaitForTopics` from https://github.com/ros2/launch_ros/pull/274 here.

Either way, we should add a comment TODO here to move the implementation back to launch_ros after https://github.com/ros2/rclpy/issues/831 is resolved.

adityapande-1995: Moved entire implementation, added a TODO comment

PR #330, opened 2021-11-09
Update python nodes sigint/sigterm handling

ivanpauno:
Related to ros2/rclpy#844.


hidmic (approved):
LGTM pending green CI

PR #337, opened 2022-02-09
Use `const&` signature for read-only sub callbacks

aprotyas:
The constant reference message signatures should be preferred over the
shared pointer to constant message signatures for read-only subscriber
callbacks, as discussed in https://github.com/ros2/rclcpp/pull/1598.

As such, it makes sense to migrate to said signatures in the examples.

Note that this should be backported to Galactic _only_.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/rclcpp/topics/minimal_subscriber/member_function.cpp
+++ b/rclcpp/topics/minimal_subscriber/member_function.cpp
@@ -31,9 +31,9 @@ public:
   }
 
 private:
-  void topic_callback(const std_msgs::msg::String::ConstSharedPtr msg) const
+  void topic_callback(const std_msgs::msg::String & msg) const
   {
-    RCLCPP_INFO(this->get_logger(), "I heard: '%s'", msg->data.c_str());
+    RCLCPP_INFO(this->get_logger(), "I heard: '%s'", msg.data.c_str());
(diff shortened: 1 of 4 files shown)


aprotyas:
CI
Repos file: https://gist.github.com/aprotyas/dca8fd39a17abacc898fa569e6b50fc5
Build args: `--packages-up-to examples_rclcpp_minimal_subscriber`
Test args: `--packages-select examples_rclcpp_minimal_subscriber`
Job: https://ci.ros2.org/job/ci_launcher/9790/
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16121/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10751/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16450/)

wjwwood (approved):
lgtm

This isn't strictly required (there's no problem using `::ConstSharedPtr` in this case), but I do agree that `const &` is easier to read and understand and also has no drawback as a recommendation. So 👍 from me.

adityapande-1995 (approved):

PR #348, opened 2022-08-02
Type adaptation examples with hardware accelerated pipelines

hemalshahNV:
There are three packages as follows:

example_type_adapters : This package implements a custom user defined ROS type for an image (type_adaptation::example_type_adapters::ImageContainer).

julia_set: An example that computes [Julia Set](https://en.wikipedia.org/wiki/Julia_set) on an incoming image.
simple_increment: A trivial example that increases each pixel value by 1 on an incoming image.

Examples show a GPU (HW accelerator) pipeline that is performing CUDA operations in a chain of intra-process nodes optimized to reduce unnecessary memory copy during message transport and eliminate unnecessary GPU to CPU synchronization.


clalancette:
First of all, thanks for opening this.

That said, there are a number of problems with this PR as it stands currently:

1. It targets the `humble` branch, and we usually put development on the `rolling` branch first.
2. We would need to fix up the Signed-off-by lines for the DCO bot to pass.
3. You'll notice that automatic CI is failing.  That's because the `cuda` package isn't available as one of our rosdep keys.

All of that said, we think that these examples are worthwhile to have, but we also don't think we should have them in the core.  They are vendor-specific and will cause our CI times (which are already very high) to go even higher.

So our proposal here is that we host these examples elsewhere.  Since these examples are NVIDIA/CUDA specific, we'd recommend that we keep these as packages inside of the `NVIDIA-ISAAC-ROS` organization under a new repository (`cuda_demos`, maybe?).  Then we can add in some Rolling CI via [action-ros-ci](https://github.com/ros-tooling/action-ros-ci).

If you don't want to host them in `NVIDIA-ISAAC-ROS`, then we can look around for another place to host them.

Thoughts?

Issue #344, opened 2022-05-30
examples_rclcpp_minimal_subscriber not building


arielc-brillianetor:
Hi,
Following the ROS2 humble tutorial for windows, all of the packages successfully build (if you build them one by one) except for the **examples_rclcpp_minimal_subscriber**. I get an error that the lib file is not found. I opened the folder and and .exe exists but no .lib file. 
Is this a problem with the CMakeList.txt file?
Thanks a lot for your help,
Ariel

C:\ros_ws\ros2_example_ws>colcon build --symlink-install --merge-install --event-handlers console_cohesion+  --packages-select examples_rclcpp_minimal_subscriber
Starting >>> examples_rclcpp_minimal_subscriber
--- output: examples_rclcpp_minimal_subscriber
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19044.
-- Found ament_cmake: 1.4.0 (C:/ros2_humble/share/ament_cmake/cmake)
-- Override CMake install command with custom implementation using symlinks instead of copying resources
-- Found rclcpp: 16.2.0 (C:/ros2_humble/share/rclcpp/cmake)
-- Found rosidl_generator_c: 3.2.0 (C:/ros2_humble/share/rosidl_generator_c/cmake)
-- Found rosidl_adapter: 3.2.0 (C:/ros2_humble/share/rosidl_adapter/cmake)
-- Found rosidl_generator_cpp: 3.2.0 (C:/ros2_humble/share/rosidl_generator_cpp/cmake)
-- Using all available rosidl_typesupport_c: rosidl_typesupport_fastrtps_c;rosidl_typesupport_introspection_c
-- Using all available rosidl_typesupport_cpp: rosidl_typesupport_fastrtps_cpp;rosidl_typesupport_introspection_cpp
-- Found rmw_implementation_cmake: 6.3.0 (C:/ros2_humble/share/rmw_implementation_cmake/cmake)
-- Found rmw_fastrtps_cpp: 6.3.0 (C:/ros2_humble/share/rmw_fastrtps_cpp/cmake)
-- Using RMW implementation 'rmw_fastrtps_cpp' as default
-- Found rclcpp_components: 16.0.1 (C:/ros_ws/dev_ws/install/share/rclcpp_components/cmake)
-- Found std_msgs: 4.5.0 (C:/ros2_humble/share/std_msgs/cmake)
[...]

PR #349, opened 2022-08-25
Windows build errors : static link wait_set_subscriber_library

adityapande-1995:
This PR fixes the problem mentioned here : https://github.com/ros2/examples/issues/344
Windows builds are not able to find ``wait_set_subscriber_library.lib``

Signed-off-by: Aditya <aditya050995@gmail.com>

### DIFF
--- a/rclcpp/topics/minimal_subscriber/CMakeLists.txt
+++ b/rclcpp/topics/minimal_subscriber/CMakeLists.txt
@@ -36,7 +36,7 @@ ament_target_dependencies(subscriber_not_composable rclcpp std_msgs)
 add_executable(subscriber_content_filtering content_filtering.cpp)
 ament_target_dependencies(subscriber_content_filtering rclcpp std_msgs)
 
-add_library(wait_set_subscriber_library SHARED
+add_library(wait_set_subscriber_library STATIC
     wait_set_subscriber.cpp
     static_wait_set_subscriber.cpp
     time_triggered_wait_set_subscriber.cpp)


adityapande-1995:
Windows CI : [Build Status](https://ci.ros2.org/job/ci_windows/17596/)

rclcpp/topics/minimal_subscriber/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ ament_target_dependencies(subscriber_not_composable rclcpp std_msgs)
 add_executable(subscriber_content_filtering content_filtering.cpp)
 ament_target_dependencies(subscriber_content_filtering rclcpp std_msgs)
 
-add_library(wait_set_subscriber_library SHARED
```
sloretz, line 39: `rclcpp_components_register_node` below is registering this as a plugin. This can [only be used with shared libraries](https://github.com/ros2/rclcpp/blob/3d69031d2a614cf02dc10ea5db153471ea32b1f2/rclcpp_components/cmake/rclcpp_components_register_node.cmake#L21-L22). I don't think the problem in #344 is this library being a shared library; instead, I think there's a problem is in whatever is expecting it not to be one.

sloretz (changes requested):

clalancette:
This doesn't look the right fix for this issue, so closing this out for now.

songyuc:
As a beginner in ROS2, I wonder if this problem has been solved yet?

Issue #340, opened 2022-04-18
`check_multiple_nodes_launch_test` test flaky on Windows


Blast545:
## Bug report

**Required Info:**

- Operating System:
  - Windows
- Installation type:
  - Source, nightly_win_deb and nightly_win_rep buildfarms
- Version or commit hash:
  - main
- DDS implementation:
  - N/A
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
This is a flaky test. It would fail sometimes when running the buildfarm jobs. 
https://ci.ros2.org/view/nightly/job/nightly_win_deb/2331/testReport/junit/launch_testing_examples.launch_testing_examples/check_multiple_nodes_launch_test/launch_testing_examples_check_multiple_nodes_launch_test/

https://github.com/ros2/examples/blob/master/launch_testing/launch_testing_examples/launch_testing_examples/check_multiple_nodes_launch_test.py

There's a "TO-DO" that may require being addressed: https://github.com/ros2/examples/blob/7d6f14a55dd3d6498fd7a8d5a90cd482ec71970c/launch_testing/launch_testing_examples/launch_testing_examples/check_multiple_nodes_launch_test.py#L83

#### Expected behavior
Test would pass.

#### Actual behavior
Test fails.

#### Additional information

This error failed for the first time in the nightly_win_deb farm: https://ci.ros2.org/view/nightly/job/nightly_win_deb/2272/ and has failed 6times counting that.

This is the first time the error showed in the repeated jobs: https://ci.ros2.org/view/nightly/job/nightly_win_rep/2505
and has failed 46 times since then. Pretty reliable in the repeated jobs.

sloretz:
I'm unable to reproduce this issue on Windows.

Crola1702:
It appeared again today: https://ci.ros2.org/view/nightly/job/nightly_win_deb/2525/

Flakiness ratio:
* 21% (15 days)
* 23% (30 days)

Log output:
<details>

```
FAIL: launch_testing_examples.check_multiple_nodes_launch_test.CheckMultipleNodesLaunched.test_node_does_not_exist
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\ci\ws\src\ros2\examples\launch_testing\launch_testing_examples\launch_testing_examples\check_multiple_nodes_launch_test.py", line 73, in test_node_does_not_exist
[...]

PR #350, opened 2022-10-26
Increase the WaitForNode timeout.

clalancette:
1 second is far too short, particularly on Windows.  This may reduce the flakiness in this test.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This may fix (or at least improve) #340 .  @Crola1702 FYI


Crola1702 (approved):
LGTM

We will keep track of this test in windows builds.

PR #357, opened 2023-03-10
Enable document generation using rosdoc2 for ament_python pkgs

Yadunund:
Fix missing `exec_depends` in `package.xmls` and docstrings to generate documentation without any warnings using `autodoc`

### DIFF
--- a/launch_testing/launch_testing_examples/package.xml
+++ b/launch_testing/launch_testing_examples/package.xml
@@ -25,14 +25,8 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
-  <test_depend>launch</test_depend>
-  <test_depend>launch_ros</test_depend>
-  <test_depend>launch_testing</test_depend>
-  <test_depend>launch_testing_ros</test_depend>
-  <test_depend>rclpy</test_depend>
   <test_depend>ros2bag</test_depend>
   <test_depend>rosbag2_transport</test_depend>
-  <test_depend>std_msgs</test_depend>
   <test_depend>demo_nodes_cpp</test_depend>
 
   <export>
--- a/launch_testing/launch_testing_examples/package.xml
+++ b/launch_testing/launch_testing_examples/package.xml
@@ -13,21 +13,21 @@
   <author email="aditya.pande@openrobotics.org">Aditya Pande</author>
   <author email="sloretz@openrobotics.org">Shane Loretz</author>
 
+  <exec_depend>demo_nodes_cpp</exec_depend>
   <exec_depend>launch</exec_depend>
   <exec_depend>launch_ros</exec_depend>
   <exec_depend>launch_testing</exec_depend>
   <exec_depend>launch_testing_ros</exec_depend>
+  <exec_depend>python3-pytest</exec_depend>
   <exec_depend>rclpy</exec_depend>
   <exec_depend>rcl_interfaces</exec_depend>
+  <exec_depend>ros2bag</exec_depend>
+  <exec_depend>rosbag2_transport</exec_depend>
   <exec_depend>std_msgs</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
-  <test_depend>python3-pytest</test_depend>
-  <test_depend>ros2bag</test_depend>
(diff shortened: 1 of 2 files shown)


launch_testing/launch_testing_examples/package.xml
```diff
@@ -13,6 +13,14 @@
   <author email="aditya.pande@openrobotics.org">Aditya Pande</author>
   <author email="sloretz@openrobotics.org">Shane Loretz</author>
 
+  <exec_depend>launch</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
+  <exec_depend>launch_testing</exec_depend>
+  <exec_depend>launch_testing_ros</exec_depend>
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend>rcl_interfaces</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
```
clalancette, line 22: package.xml should not contain both an `exec_depend` and a `test_depend` for the same dependency; see http://wiki.ros.org/catkin/package.xml section 3.3.

In this case, it looks like all of these really should be `exec_depend`, not `test_depend`, so I'd suggest removing them from the list below.

clalancette (changes requested):

launch_testing/launch_testing_examples/package.xml
Yadunund: Thanks for enlightening me about the manifest requirement!
I've removed the `test_depends` [70e2453](https://github.com/ros2/examples/pull/357/commits/70e2453a8b8dd13951c9ba6b0ef10f253a18ebe4)

launch_testing/launch_testing_examples/package.xml
```diff
@@ -13,18 +13,20 @@
...
+
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
```
clalancette, line 27: This should also be an `exec_depend`, not a `test_depend`.

launch_testing/launch_testing_examples/package.xml
```diff
@@ -13,18 +13,20 @@
...
-  <test_depend>launch</test_depend>
-  <test_depend>launch_ros</test_depend>
-  <test_depend>launch_testing</test_depend>
-  <test_depend>launch_testing_ros</test_depend>
-  <test_depend>rclpy</test_depend>
   <test_depend>ros2bag</test_depend>
   <test_depend>rosbag2_transport</test_depend>
```
clalancette, line 29: Both of these should be `exec_depend` as well.

launch_testing/launch_testing_examples/package.xml
```diff
@@ -13,18 +13,20 @@
...
-  <test_depend>std_msgs</test_depend>
   <test_depend>demo_nodes_cpp</test_depend>
```
clalancette, line 30: This should also be an `exec_depend`.

clalancette (changes requested):

launch_testing/launch_testing_examples/package.xml
Yadunund: Gotcha, made all the changes [496e469](https://github.com/ros2/examples/pull/357/commits/496e4693f7255eaa2303b986f4a7139989ca3d99)

launch_testing/launch_testing_examples/package.xml
Yadunund: To clarify, `python3-pytest` seems to be listed as `test_depend` in several other `ament_python` packages in `ros2` org.

clalancette: Yeah.  The difference in this package is that `pytest` is used in the actual code of the package, not in the tests (that is, the code that does `import pytest` is in the `launch_testing_examples` subdirectory, not the `test` subdirectory).  Thus, these should all be `exec_depend`.

clalancette (approved):
Thanks for the iteration!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18226/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12757/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18895/)

launch_testing/launch_testing_examples/package.xml
Yadunund: Got it 👍🏼

PR #362, opened 2023-05-20
Add test package for single and multi thread

icsl-Jeon:
Hello. My name is Felipe Jeon, and I am glad to open a PR regarding comparisons on SingleThreadedExecutor and MultiThreadedExecutor (reentrant and mutually exclusive).
I attached detailed analysis on the [link](https://motion-boseong.vercel.app/threading-ros2) inside README.md.


clalancette (changes requested):
First of all, thanks for submitting this and sorry that it has taken so long to look into it.

I have a few comments around the structure of this PR:
1.  In the README, please don't point to an external post for documentation about the threading.  If there is something we want to explain about how this works, it should be done right here in the README (or possibly on https://docs.ros.org, but I think here in the README makes more sense).
2. All of the files need license and copyright headings, like in https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/multithreaded_executor/multithreaded_executor.cpp
3. I'm not totally sure of the structure of this, and how it fits in with our other executor examples.  That is, how much does this overlap with https://github.com/ros2/examples/tree/rolling/rclcpp/executors/multithreaded_executor ?  Maybe we should think about combining these into the same package there?
4. Although we aren't totally consistent about this, our style is make class names CamelCase and to make function/method names snake_case.
5. Our class member variables are always suffixed with a `_` (that's the case in some of this code, but not all of it).
6. For installation, in general the installation directory should match the name of the package, not a different name.
7. A license and a package description need to be filled in in the package.xml

PR #363, opened 2023-05-29
Split lambda and subscription definition in minimal subscriber example

FelipeGdM:
The current example of minimal subscriber using a lambda function performs the definition of the subscription and the lambda function in the same line.

This PR splits the same operation in two steps, the definition of the lambda function and then the definition of the subscription. The goal is make it easier to explain what is happening in each step, as this example will be the default one in the ROS 2 documentation page (see https://github.com/ros2/ros2_documentation/issues/3618). Another benefit is that the minimal subscriber with lambdas would now follow the same structure of the minimal publisher with lambdas (i.e. declaration of the lambda and then declaration of the publisher/subscriber)

The name of the lambda function variable is `topic_callback` in order to match the same pattern from the minimal subscriber member function example

### DIFF
--- a/rclcpp/topics/minimal_subscriber/lambda.cpp
+++ b/rclcpp/topics/minimal_subscriber/lambda.cpp
@@ -23,12 +23,12 @@ public:
   MinimalSubscriber()
   : Node("minimal_subscriber")
   {
-    subscription_ = this->create_subscription<std_msgs::msg::String>(
-      "topic",
-      10,
-      [this](std_msgs::msg::String::UniquePtr msg) {
+    auto topic_callback =
+      [this](std_msgs::msg::String::UniquePtr msg) -> void {
         RCLCPP_INFO(this->get_logger(), "I heard: '%s'", msg->data.c_str());
-      });
+      };
+    subscription_ =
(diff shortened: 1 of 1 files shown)


ahcorde (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18834/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13362/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19578/)

clalancette:
I'm fine with this change, but we should check the documentation at https://github.com/ros2/ros2_documentation to see if we need to update any of the Tutorials because of this.

clalancette:
> @clalancette this particular example isn't used in the ROS 2 docs, the current example is the [member_function.cpp](https://github.com/ros2/examples/blob/rolling/rclcpp/topics/minimal_subscriber/member_function.cpp). This change is part of the effort to replace `std::bind` calls in the default examples in favor of lambda functions ([ros2/ros2_documentation#3618](https://github.com/ros2/ros2_documentation/issues/3618)), so it should be used in the docs in the near future.

Sounds good, thanks!

PR #369, opened 2024-01-10
Fix foxy tutorial link

tfoote:
Fixes: https://github.com/ros-infrastructure/rosindex/issues/336

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
 ROS 2 examples
 ==============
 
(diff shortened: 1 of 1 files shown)


clalancette:
I'll just point out that there are hundreds of these broken URLs across all of our EOL distributions.  I don't think in general we should set out to fix them.  I don't think we should revert this, but I also don't think we should do any more of these.

PR #373, opened 2024-03-04
Refactor WaitForNodes class.

clalancette:
The main goal here is to get rid of the sometimes
flaky tests we are seeing in the nightlies.  To do that, we try increasing the default timeout from 10 to 15 seconds.

While we are in here, do a few other pieces of refactoring in this test:

1. Use a common variable name for the node name.
2. Switch the variable name for the loop to 'finished', which better reflects what it is doing.
3. Change the sleep between loops to 0.1, so we react faster
4. Remove the unnecessary _prepare_node() method.

I think this should make the flaky test as in https://ci.ros2.org/view/nightly/job/nightly_linux_debug/2978/testReport/junit/launch_testing_examples.launch_testing_examples/check_multiple_nodes_launch_test/launch_testing_examples_check_multiple_nodes_launch_test/ less flaky.

### DIFF
--- a/launch_testing/launch_testing_examples/launch_testing_examples/check_multiple_nodes_launch_test.py
+++ b/launch_testing/launch_testing_examples/launch_testing_examples/check_multiple_nodes_launch_test.py
@@ -28,57 +28,6 @@ import rclpy
 from rclpy.node import Node
 
 
-@pytest.mark.launch_test
-@launch_testing.markers.keep_alive
-def generate_test_description():
-    launch_actions = []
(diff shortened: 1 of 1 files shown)


ahcorde (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20419/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14821/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21161/)

PR #374, opened 2024-03-08
Cleanup the launch_testing_examples.

clalancette:
This is to make them more robust when run in parallel with other tests.  In particular, we make sure that the tests cleanup after themselves, and we also increase the timeouts to 15 seconds.

This should improve some of the flaky tests in this package, like we see in https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/3385/#showFailuresLink

### DIFF
--- a/launch_testing/launch_testing_examples/launch_testing_examples/set_param_launch_test.py
+++ b/launch_testing/launch_testing_examples/launch_testing_examples/set_param_launch_test.py
@@ -48,6 +48,7 @@ if os.name != 'nt':
             self.node = Node('test_node')
 
         def tearDown(self):
+            self.node.destroy_node()
             rclpy.shutdown()
 
         def test_set_parameter(self, proc_output):
--- a/launch_testing/launch_testing_examples/launch_testing_examples/set_param_launch_test.py
+++ b/launch_testing/launch_testing_examples/launch_testing_examples/set_param_launch_test.py
@@ -43,26 +43,25 @@ def generate_test_description():
 if os.name != 'nt':
     class TestFixture(unittest.TestCase):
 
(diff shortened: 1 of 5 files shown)


launch_testing/launch_testing_examples/launch_testing_examples/set_param_launch_test.py
```diff
@@ -43,26 +43,25 @@ def generate_test_description():
...
+        def setUp(self):
             rclpy.init()
-            node = Node('test_node')
-            response = set_parameter(node, value=True)
-            assert response.successful, 'Could not set parameter!'
-            rclpy.shutdown()
+            self.node = Node('test_node')
 
+        def tearDown(self):
+            rclpy.shutdown()
```
fujitatomoya, line 51: do we need `destroy_node` here?

```suggestion
            self.node.destroy_node()
            rclpy.shutdown()
```

clalancette: Yes! 100%. That was part of what I was trying to fix here, I can't believe I forgot it. Thanks for pointing it out.

fujitatomoya (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20451/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14853/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21194/)

clalancette:
I believe that these fixes will also fix up #391, so I'll backport this to Humble and Iron.

PR #377, opened 2024-04-30
Switching to examples_interfaces

### LINKED ISSUE #356: Official ros2 examples still use deprecated msg?
std_msgs have been deprecated in https://github.com/ros2/common_interfaces/pull/116; why are they still being [used](https://github.com/ros2/examples/blob/b327f8cd19a8278631116f01b96bdfc7b55205ec/rclpy/topics/minimal_publisher/examples_rclpy_minimal_publisher/publisher_local_function.py#L17) in this package?

It's hard for users to make the switch when official examples still use them

CursedRock17:
Meant to resolve #356, this pull request switches all the deprecated instances of `std_msgs` to `example_interfaces` after deprecation of the many elements occured. Though, any msg with `Header` is still a `std_msgs` since they were not part of that deprecation.


ahcorde (approved):

CursedRock17:
Hmm looks like a regression with one test, do you think it needed to be patched in #374? I can't imagine changing to `example_interfaces` would alter the time it takes to deliver [one specific message](https://github.com/ros2/examples/blob/6affb93eee6206a6411a0b065df1ff6b6a5b7756/launch_testing/launch_testing_examples/launch_testing_examples/check_msgs_launch_test.py#L43), but I could be wrong.

fujitatomoya:
@CursedRock17 https://github.com/ros2/examples/pull/374 is already included in your repo https://github.com/CursedRock17/examples/tree/deprecated_msgs? what do you mean by https://github.com/ros2/examples/pull/377#issuecomment-2086093785?

CursedRock17:
Yeah, I just didn't know if that specific test had passed all the way through, I'm trying to figure out why that specific test continues to be a regression. I guess it did pass [in #374's CI](https://ci.ros2.org/job/ci_linux-aarch64/14853/testReport/launch_testing_examples.launch_testing_examples/check_msgs_launch_test/) so my changes must have broke it somehow.

launch_testing/launch_testing_examples/launch_testing_examples/check_msgs_launch_test.py
```diff
@@ -14,14 +14,14 @@
 
 import unittest
 
+from example_interfaces.msg import String
```
fujitatomoya, line 17: i believe this changes led to https://ci.ros2.org/job/ci_linux/20965/testReport/junit/launch_testing_examples.launch_testing_examples/check_msgs_launch_test/launch_testing_examples_check_msgs_launch_test/.

this is because message type does not match anymore,

`demo_nodes_cpp` uses `std_msgs/String`, so i think this change requires the same change to https://github.com/ros2/demos

CursedRock17:
Ok I understand it's the changes I've made that cause downstream CI failures, but I just don't see the correlation between this repo and something like [rmw_implementation](https://github.com/ros2/rmw_implementation/blob/rolling/test_rmw_implementation/CMakeLists.txt) which is currently failing. Since neither repo includes each other I assumed it could've been [`test_msgs`](https://github.com/ros2/rmw_implementation/blob/rolling/test_rmw_implementation/CMakeLists.txt#L27) but again there's no generation of examples. Could somebody point out what I'm missing when it comes to these failures? Thanks.

PR #379, opened 2024-05-03
Update the shutdown handling in all of the Python examples.

### LINKED ISSUE #372: Update python nodes SIGINT handling
## Bug report

`rclpy` was updated to automatically handle SIGINT as part of this PR which was incorporated into the Humble release: https://github.com/ros2/rclpy/pull/844

The python nodes in the demos repository were subsequently updated via the following PRs, but those changes were never incorporated into the examples repository.

* https://github.com/ros2/demos/pull/539
* https://github.com/ros2/demos/pull/555
* https://github.com/ros2/demos/pull/581

For consistency, and to ensure that an exception is not thrown when SIGINT is passed to the example python nodes, these same changes should be incorporated into this repository.

clalancette:
In particular, we should deal with KeyboardInterrupt (by just doing 'pass'), as well as ExternalShutdownException (which can come from the executor).  In all cases, we should call 'rclpy.try_shutdown()' to cleanup at the end.

Note well that I removed some of the *destroy() calls in the cleanup.  That's because they weren't correct in all cases, and to fix them up properly would really require us to have a nested set of try..except statements.  Given that these are examples, having that complex set of exception handling didn't seem like the correct way to go here.

This should fix #372 .  @fujitatomoya FYI

### DIFF
--- a/rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client.py
+++ b/rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client.py
@@ -83,8 +83,6 @@ def main(args=None):
         pass
     except ExternalShutdownException:
         sys.exit(1)
-    finally:
-        rclpy.try_shutdown()
 
 
 if __name__ == '__main__':
--- a/rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client.py
+++ b/rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client.py
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 from action_msgs.msg import GoalStatus
 from example_interfaces.action import Fibonacci
 
 import rclpy
 from rclpy.action import ActionClient
+from rclpy.executors import ExternalShutdownException
 from rclpy.node import Node
 
 
@@ -70,11 +73,18 @@ class MinimalActionClient(Node):
 def main(args=None):
     rclpy.init(args=args)
 
-    action_client = MinimalActionClient()
+    try:
+        action_client = MinimalActionClient()
 
-    action_client.send_goal()
+        action_client.send_goal()
 
-    rclpy.spin(action_client)
+        rclpy.spin(action_client)
+    except KeyboardInterrupt:
+        pass
+    except ExternalShutdownException:
+        sys.exit(1)
+    finally:
+        rclpy.try_shutdown()
 
 
 if __name__ == '__main__':
--- a/rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client_cancel.py
+++ b/rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client_cancel.py
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 from example_interfaces.action import Fibonacci
 
 import rclpy
 from rclpy.action import ActionClient
+from rclpy.executors import ExternalShutdownException
 from rclpy.node import Node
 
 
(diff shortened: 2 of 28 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20987/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15289/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21668/)

fujitatomoya (approved):
lgtm.

rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client.py
```diff
@@ -70,11 +73,18 @@ def send_goal(self):
...
+        action_client.send_goal()
 
-    rclpy.spin(action_client)
+        rclpy.spin(action_client)
+    except KeyboardInterrupt:
+        pass
+    except ExternalShutdownException:
+        sys.exit(1)
+    finally:
+        rclpy.try_shutdown()
```
wjwwood, line 87: I know this is just following up on more changes due to https://github.com/ros2/rclpy/pull/830, but it seems like a real shame to me that it's not just:

```suggestion
        rclpy.spin(action_client)
    except KeyboardInterrupt:
        pass
```

It's a discussion for another time (maybe never given the list of things we have to do), but I don't see why the executor is throwing when shutdown (externally or otherwise):

https://github.com/ros2/rclpy/blame/99f1ce933ddd1677fda8fa602af3684b93f814eb/rclpy/rclpy/executors.py#L689C5-L689C5

In rclcpp the executor just stops and spin returns without exception. Seems like a reasonable behavior for python too, but maybe I'm missing some context.

wjwwood: Also, I'm not sure why SIGTERM (or actually any shutdown of the context) results in `sys.exit(1)`.

wjwwood (approved):
Actual changes lgtm, but the thing we're propgating here is suboptimal, at least in my opinion.

rclpy/actions/minimal_action_client/examples_rclpy_minimal_action_client/client.py
wjwwood: I talked to @sloretz about it off-line, and he doesn't remember a strong reason why shutting down the executor or the context should cause spin to throw. In my opinion it's an unnecessary deviation from how rclcpp works. I don't know if @ivanpauno is watching, but maybe he has some insight as well.

Either way, @sloretz and I both agreed that the `finally: try_shutdown()` stuff can just go, again unless someone else has a good reason for it.

clalancette: I agree with you completely; the fact that we have to do this whole dance seems incorrect to me.  Given the way that Python handles `KeyboardInterrupt`, I think we *have* to do that bit, but the rest of it seems to me like we aren't doing signal handling properly.  This also manifests in the call to `rclpy.try_shutdown()`; in my opinion, that shouldn't exist at all and `rclpy.shutdown()` should just do whatever the correct thing is (even if that means deferring shutdown because something is in use).

All of that said, I think for now we should put this in to make these examples a little quieter, and then address the signal handling separately.

clalancette: > Either way, @sloretz and I both agreed that the `finally: try_shutdown()` stuff can just go, again unless someone else has a good reason for it.

Hm.  All of our rclcpp examples use the pattern:

```
rclcpp::init();
do_stuff();
rclcpp::shutdown();
```

In rclpy, the equivalent would be:

```
rclpy.init()
try:
  do_stuff()
finally:
  rclpy.try_shutdown()
```

I don't really feel that strongly about it, but it does kind of feel like we should do the shutdown in rclpy as well, no?

clalancette: All right.  After further discussion, we agreed to remove the `try_shutdown` everywhere to make the examples simpler.  We should also remove `ExternalShutdownException`, but that will be done at a later time; see #1279.

We also discussed that this would all be a lot nicer if the various ROS 2 entities in rclpy were context managers, so I've also opened up #1280 to talk about that.

With that, I've now done all of the updates for this PR, so I could use new reviews on this.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21012/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15316/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21693/)

wjwwood (approved):

clalancette:
Going ahead and merging this one in.

PR #382, opened 2024-05-16
Use a single executor instance for spinning in client_async_callback.

clalancette:
In the current code, it calls rclpy.spin_once(), which instantiates a new executor, adds the node to it, executors one work item, then removes the node and destroys the executor.

Besides being inefficient, the problem with that sequence is that adding a node to the executor actually ends up being an "event", and so the work item that gets returned can be just the work of adding the node, over and over again. For reasons I admit I don't fully understand, this only happens with rmw_cyclonedds_cpp, not with rmw_fastrtps_cpp.

Regardless, the much more performant thing to do is to create an executor at the beginning of the example and to just spin on that.  This eliminates adding it to the node constantly, and makes this work with all RMWs.

### DIFF
--- a/rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async_callback.py
+++ b/rclpy/services/minimal_client/examples_rclpy_minimal_client/client_async_callback.py
@@ -26,9 +26,13 @@ def main(args=None):
 
     try:
         node = rclpy.create_node('minimal_client')
+
+        executor = rclpy.executors.SingleThreadedExecutor()
+        executor.add_node(node)
+
         # Node's default callback group is mutually exclusive. This would prevent the client
         # response from being processed until the timer callback finished, but the timer callback
-        # int this example is waiting for the client response
+        # in this example is waiting for the client response
         cb_group = ReentrantCallbackGroup()
         cli = node.create_client(AddTwoInts, 'add_two_ints', callback_group=cb_group)
         did_run = False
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
the fix lgtm, but i cannot explain why this is happening with cyclonedds either.

it never takes the service response from the server, server has sent the service response to the client for sure. (maybe we would keep this follow-up on rmw_cyclonedds repo. it is easy to reproduce to revert this fix.)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21096/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15398/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21812/)

clalancette:
See https://github.com/ros2/rmw_cyclonedds/issues/494

wjwwood (approved):
lgtm

I am digging into why this happens only with rmw_cyclonedds. As you pointed out in https://github.com/ros2/rmw_cyclonedds/issues/494 @clalancette, it's due to a single kind of event being processed each time you spin which is never the event you want.

I think this change is just better though. The "global" version of `spin_once()` is kind of not that useful and in this case harmful, at least in my opinion. We could consider deprecating and then removing it. But the issue with cyclone should still be followed up on.

wjwwood:
@ros-pull-request-builder retest this please

wjwwood:
I think the Rpr failure is an unrelated (probably pre-existing) failure: https://build.ros2.org/job/Rpr__examples__ubuntu_noble_amd64/19/testReport/launch_testing_examples.launch_testing_examples/record_rosbag_launch_test/launch_testing_examples_record_rosbag_launch_test/

clalancette:
> I think the Rpr failure is an unrelated (probably pre-existing) failure: https://build.ros2.org/job/Rpr__examples__ubuntu_noble_amd64/19/testReport/launch_testing_examples.launch_testing_examples/record_rosbag_launch_test/launch_testing_examples_record_rosbag_launch_test/

Yeah, agreed.  I'm going to go ahead and merge this in, thanks for the reviews!

clalancette:
@Mergifyio backport jazzy

PR #432, opened 2025-12-23
Modify timer callback execution for ROS2 Humble

michaelakpore:
Title:
Fix Humble API: execute_callback() no longer takes arguments

Description:
- In ROS2 Humble, TimerBase::execute_callback() no longer accepts any arguments.
- The example time_triggered_wait_set_subscriber.cpp previously called
  execute_callback(data), which fails to compile.
- This change updates the example to use execute_callback() with no arguments,
  and adds a comment above the line for future readers:

    // NOTE: In ROS2 Humble, TimerBase::execute_callback() no longer takes arguments

Tested by building with colcon and running the subscriber/publisher example.

Is this user-facing behavior change? No
Did you use Generative AI? No
Additional Information: Tested on ROS2 Humble LTS, Ubuntu 22.04.

### DIFF
--- a/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
+++ b/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
@@ -71,8 +71,10 @@ public:
         case rclcpp::WaitResultKind::Ready:
           {
             if (wait_result.get_wait_set().get_rcl_wait_set().timers[0U]) {
-              if (auto data = timer_->call()) {
-                timer_->execute_callback(data);
+              // NOTE: In ROS2 Humble, TimerBase::execute_callback() no longer takes arguments.
+              // This replaces the old execute_callback(data) call from Foxy and earlier.
+              if (timer_->call()) {
+                timer_->execute_callback();
               }
             }
             break;
--- a/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
+++ b/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
@@ -74,7 +74,7 @@ public:
               // NOTE: In ROS2 Humble, TimerBase::execute_callback() no longer takes arguments.
               // This replaces the old execute_callback(data) call from Foxy and earlier.
               if (auto data = timer_->call()) {
-              #if RCLCPP_VERSION_MAJOR >= 16
+              #if RCLCPP_VERSION_MAJOR <= 16
                 timer_->execute_callback();
                 #else
                 timer_->execute_callback(data);
--- a/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
+++ b/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
@@ -73,8 +73,12 @@ public:
             if (wait_result.get_wait_set().get_rcl_wait_set().timers[0U]) {
               // NOTE: In ROS2 Humble, TimerBase::execute_callback() no longer takes arguments.
               // This replaces the old execute_callback(data) call from Foxy and earlier.
-              if (timer_->call()) {
+              if (auto data = timer_->call()) {
+                #if RCLCPP_VERSION >= RCLCPP_VERSION_COMBINED(Humble)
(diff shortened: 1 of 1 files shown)


rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
```diff
@@ -71,8 +71,14 @@ class TimeTriggeredWaitSetSubscriber : public rclcpp::Node
         case rclcpp::WaitResultKind::Ready:
           {
             if (wait_result.get_wait_set().get_rcl_wait_set().timers[0U]) {
+              // NOTE: In ROS2 Humble, TimerBase::execute_callback() no longer takes arguments.
+              // This replaces the old execute_callback(data) call from Foxy and earlier.
```
fujitatomoya, line 75: Humble or earlier distro (which are already End of Life, those are not maintained by ROS PMC anymore), are implemented in this way as examples. so it does not need to do API negotiation like this.

rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp
```diff
@@ -71,8 +71,14 @@ class TimeTriggeredWaitSetSubscriber : public rclcpp::Node
...
               if (auto data = timer_->call()) {
+              #if RCLCPP_VERSION_MAJOR >= 16
```
fujitatomoya: i do not think this is correct. Humble major version generated by ament is already 16, that means this includes Humble or later release. probably what you are trying to do is that,


```suggestion
              #if RCLCPP_VERSION_MAJOR <= 16
```

but i do not think this matters...

fujitatomoya (commented):
> In ROS2 Humble, TimerBase::execute_callback() no longer accepts any arguments.

that is how implemented for humble: https://github.com/ros2/examples/blob/09899ca6b6361f2cca6d4f626c0304c86e7b9ada/rclcpp/topics/minimal_subscriber/time_triggered_wait_set_subscriber.cpp#L74

> The example time_triggered_wait_set_subscriber.cpp previously called execute_callback(data), which fails to compile.

i think that is because you are using examples from rolling branch on humble workspace to build. this is not expected procedure.

fujitatomoya:
@michaelakpore thanks for pushing the fix.

before you are making more effort on this as code base, i would like to say again that each repository such as https://github.com/ros2/examples has multiple branches that target the distribution. so my point is that we do not have build failure in the 1st place even without this API version check. please see my comment above.

thanks,

michaelakpore:
@fujitatomoya Thanks for the clarification, and I appreciate you reiterating this.

I understand now that ros2/examples maintains per-distribution branches, and that Humble-specific APIs should already be correctly reflected in the Humble branch without needing API negotiation or conditional handling.

My intention was to fix the compilation issue I encountered when building Humble locally, but I see that using code from a different distribution branch in a Humble workspace is not an expected or supported workflow.

I’m happy to close or adjust this PR accordingly based on you're preference. Thanks for taking the time to explain this.

fujitatomoya:
@michaelakpore thanks for iterating with me.

you have already closed this PR, i think that would be probably best to do here.

by the way, i think you brought up the good point here about `RCLCPP_VERSION`, `RCL_VERSION`, `RMW_VERSION` and related macros are auto-generated and useful for API negotiation. and these are not documented anywhere in https://github.com/ros2/ros2_documentation.

Even though ROS 2 guarantees ABI compatibility within the same distribution, new interfaces and features can be backported. This means that within a single distro, there could be different API versions available. In such cases, these version check macros are really helpful for ensuring user application integrity by allowing developers to check if a new feature is available before using it.

These macros exist and are useful, but developers don't know about them because they're undocumented, limiting their ability to write robust code that works across different API versions within the same distribution. i will create the issue for https://github.com/ros2/ros2_documentation to see if this is good proposal to enhance the documentation.

fujitatomoya:
@michaelakpore JFYI, https://github.com/ros2/ros2_documentation/issues/6115

michaelakpore:
@fujitatomoya Thanks for the heads-up! I’ve taken a look at ros2/ros2_documentation#6115 — really appreciate you creating the issue and carrying this forward.

